PR #3, opened 2020-11-16
Story/Q20-2

aa-kashk:
Admin should be able to add new apps


quadriga/src/main/java/edu/asu/diging/quadriga/core/db/impl/AppDbConnection.java
```diff
@@ -0,0 +1,74 @@
...
+import edu.asu.diging.quadriga.core.exceptions.UnstorableObjectException;
+import edu.asu.diging.quadriga.core.model.IApp;
+import edu.asu.diging.quadriga.core.model.impl.App;
+
+@Component
+@Transactional
+public class AppDbConnection implements IAppDbConnection {
+
+    @PersistenceContext
+    private EntityManager em;
```
jdamerow: you should use repositories instead of the entitymanager; this class is probably not necessary at all, just use the repositories from the manager classes.

quadriga/src/main/java/edu/asu/diging/quadriga/core/db/impl/AppDbConnection.java
```diff
@@ -0,0 +1,74 @@
...
+    public String generateId() {
+        String id = null;
+        while (true) {
+            id = "APP" + generateUniqueId();
+            Object existingFile = getById(id);
+            if (existingFile == null) {
+                break;
+            }
+        }
+        return id;
```
jdamerow: check citesphere or virtual spaces  for how to create ids in this form (through annotation and a generator)

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/AppRepository.java
```diff
@@ -0,0 +1,13 @@
+package edu.asu.diging.quadriga.core.data;
+
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+import edu.asu.diging.quadriga.core.model.IApp;
+import edu.asu.diging.quadriga.core.model.impl.App;
+
+@Repository
+public interface AppRepository extends CrudRepository<App, String>{
```
jdamerow: you probably want a `PagingAndSortingRepository` repo

quadriga/src/main/java/edu/asu/diging/quadriga/core/db/impl/AppDbConnection.java
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddAppController.java
```diff
@@ -0,0 +1,51 @@
...
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    @Autowired
+    private IAppManager appManager;
+
+
+    @RequestMapping(value = "/admin/add", method=RequestMethod.GET)
+    public String get(Model model) {
+        model.addAttribute("app", new AppForm());
+        return "addApp";
```
jdamerow: since that's an admin function, the view file should be in an appropriate folder

jdamerow (changes requested):
- please remove classes that are not needed anymore

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddAppController.java
```diff
@@ -0,0 +1,51 @@
...
+@Controller
+public class AddAppController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    @Autowired
+    private IAppManager appManager;
+
+
+    @RequestMapping(value = "/admin/add", method=RequestMethod.GET)
```
jdamerow: since this is not adding an admin, i would suggest `/admin/apps/add` or something like that ;)

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Service
+@Transactional
+public class AppManager implements IAppManager {
+
+    @Autowired
+    private AppRepository appRepo;
+
+    @Override
+    public void addApp(AppForm appForm) throws UnstorableObjectException {
```
jdamerow: where is `UnstorableObjectException` being thrown?

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/exceptions/UnstorableObjectException.java
```diff
@@ -0,0 +1,33 @@
+package edu.asu.diging.quadriga.core.exceptions;
+
+public class UnstorableObjectException extends Exception {
```
jdamerow: is this class used anywhere?

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddAppController.java
```diff
@@ -0,0 +1,36 @@
...
+        return "admin/addApp";
+    }
+
+    @RequestMapping(value = "/admin/apps/add", method = RequestMethod.POST)
+    public String add(@ModelAttribute AppForm appForm, RedirectAttributes redirectAttrs) {
+
+        appManager.addApp(appForm);
+
+        redirectAttrs.addFlashAttribute("alert_type", "success");
+        redirectAttrs.addFlashAttribute("alert_msg", "App was successfully added.");
```
jdamerow: since we're starting from fresh here, let's do proper message handling with message.properties, so that here you would just return the message id.

jdamerow (changes requested):
Let's also add a list page that lists all apps so that it's easy to see what had been added.

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddAppController.java
```diff
@@ -0,0 +1,42 @@
...
+        return "admin/addApp";
+    }
+
+    @RequestMapping(value = "/admin/apps/add", method = RequestMethod.POST)
+    public String add(@ModelAttribute AppForm appForm, RedirectAttributes redirectAttrs) {
+
+        appManager.addApp(appForm);
+
+        redirectAttrs.addFlashAttribute("alert_type", "success");
+        redirectAttrs.addFlashAttribute("alert_msg", env.getRequiredProperty("101"));
```
jdamerow: this should use message properties and locales. property names should be descriptive of what they are for.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddAppController.java
```diff
@@ -0,0 +1,42 @@
...
+        return "admin/addApp";
+    }
+
+    @RequestMapping(value = "/admin/apps/add", method = RequestMethod.POST)
+    public String add(@ModelAttribute AppForm appForm, RedirectAttributes redirectAttrs) {
+
+        appManager.addApp(appForm);
+
+        redirectAttrs.addFlashAttribute("alert_type", "success");
+        redirectAttrs.addFlashAttribute("alert_msg", env.getRequiredProperty("app_create_success"));
```
jdamerow: you are still always reading the English message file. If the user is in France, they want to see the French version. This should help I think: https://medium.com/skillhive/internalization-in-spring-boot-22f4fa82f132

quadriga/src/main/webapp/WEB-INF/views/admin/addApp.html
```diff
@@ -0,0 +1,24 @@
...
+</head>
+<body>
+	<div layout:fragment="content">
+	<span th:text="${alert_msg}"></span>
+		<h2>Add new App</h2>
+
+		<form action="#" th:action="@{/admin/apps/add}" method="POST"
+			th:object="${app}">
+			<input type="hidden" name="${_csrf.parameterName}"
+				value="${_csrf.token}" />
```
jdamerow: have you checked if you actually need this? with jsps the csrf token was automatically added. not sure if it's the same with thymeleaf

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/impl/App.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hibernate.annotations.GenericGenerator;
+
+import edu.asu.diging.quadriga.core.model.IApp;
+
+@Entity
+public class App implements IApp {
+
+    @Id
+    @GeneratedValue(generator = "app_id_generator")
+    @GenericGenerator(name = "app_id_generator", parameters = @Parameter(name = "prefix", value = "app"), strategy = "edu.asu.diging.quadriga.core.data.IdGenerator")
```
jdamerow: Let's uppercase the prefix.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
```diff
@@ -0,0 +1,29 @@
...
+
+import edu.asu.diging.quadriga.core.data.AppRepository;
+import edu.asu.diging.quadriga.core.model.IApp;
+import edu.asu.diging.quadriga.core.model.impl.App;
+import edu.asu.diging.quadriga.core.service.IAppManager;
+import edu.asu.diging.quadriga.web.forms.AppForm;
+
+@Service
+@Transactional
+public class AppManager implements IAppManager {
```
jdamerow, line 15: needs javadoc and test cases

jdamerow (changes requested):

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/AppManagerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public void test_addApp_success() {
+        AppForm appForm = new AppForm();
+        String name = "name";
+        String desc = "testApp";
+        appForm.setName(name);
+        appForm.setDescription(desc);
+        App app = managerToTest.addApp(appForm);
+        Mockito.verify(appRepo).save(app);
+        Assert.assertEquals(name,app.getName());
+        Assert.assertEquals(desc,app.getDescription());
```
jdamerow: you also want to test that the app has an id.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

quadriga/src/main/webapp/WEB-INF/views/layouts/main.html
```diff
@@ -49,6 +49,9 @@
 	          	<li sec:authorize="hasRole('ADMIN')" role="presentation">
 	          		<a th:href="@{/admin/user/list}" >Users</a>
 	          	</li>
+	          	<li sec:authorize="hasRole('ADMIN')" role="presentation">
+	          		<a th:href="@{/admin/add}" >Add App</a>
```
jdamerow: Let's call this Apps

jdamerow (changes requested):
Besides those should be good to go.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/AppManagerTest.java
aa-kashk: I need to discuss this with you. The appId is being generated in the database and I am unable to return a mock for `App` instance as it is being instantiated inside the service layer. Assert returns False as app's Id during test is `null`

jdamerow: Looking at your code, there is actually a potential bug that is related to this. Your addApp method should return the saved app it gets from the repo (instead of saving and returning the app that was passed to the repository), since only that one will have the new id, which the caller might need. If you do that, then you will need to mock the save method of the repository anyways and you can use that to set an id. Does this make sense?

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/AppRepository.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Repository
+public interface AppRepository extends PagingAndSortingRepository<App, String> {
+
+    default App saveApp(AppForm appForm) {
+        App app = new App();
+        app.setName(appForm.getName());
+        app.setDescription(appForm.getDescription());
+        this.save(app);
+        return app;
```
jdamerow: this logic should go in the service/manager class;

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
```diff
@@ -0,0 +1,34 @@
...
+     * 
+     * @param appForm Form data that needs to be added to database
+     * 
+     * 
+     * @return App Instance that is saved in database
+     * 
+     **/
+    @Override
+    public App addApp(AppForm appForm) {
+        return appRepo.saveApp(appForm);
```
jdamerow: the logic you put in the repo should go here. the repo would be an empty interface for now

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
```diff
@@ -0,0 +1,39 @@
...
+     * 
+     * @return App Instance that is saved in database
+     * 
+     **/
+    @Override
+    public App addApp(AppForm appForm, App app) {
+        app.setName(appForm.getName());
+        app.setDescription(appForm.getDescription());
+        app = appRepo.save(app);
+        return app;
```
jdamerow, line 35: merge with line 35

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
```diff
@@ -0,0 +1,39 @@
...
+     * Creates a new App instance and stores it in the database
+     * 
+     * @param appForm   Form data that needs to be added to database
+     * 
+     * 
+     * @return App Instance that is saved in database
+     * 
+     **/
+    @Override
+    public App addApp(AppForm appForm, App app) {
```
jdamerow: there is really no reason to pass an `App` here is there? Since it's adding new apps, it should just create that in the method.

jdamerow (changes requested):

aa-kashk:
- I am still facing an issue creating the mock of App object if I create the App instance in the manager. 
- I am able to verify the name and the description but the Id is getting generated through MySQL. Not sure how to mock the App instance correctly.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AppManager.java
jdamerow: merge with line 35

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/AppManagerTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    public void test_addApp_success() {
+        AppForm appForm = new AppForm();
+        String name = "name";
+        String description = "testApp";
+        appForm.setName(name);
+        appForm.setDescription(description);
+        App app = managerToTest.addApp(appForm);
+        Mockito.verify(appRepo).save(app);
```
jdamerow, line 36: `save` method of appRepo needs to be mocked to return an app instance with an id. you can either just make sure the app object passed into `save` is of the right type or use and argumentMatcher or argumentCaptor if you want to go beyond that.

jdamerow (changes requested):

jdamerow:
[Q20-2] changed status to "Won't Do".

PR #5, opened 2021-03-30
story/Q20-5

aa-kashk:
Users should be able to create new collections

- Collections have name and description; they should be stored in mongodb


jdamerow:
Was there no automatic text added to the PR when you opened it?

quadriga/src/main/java/edu/asu/diging/quadriga/core/mongo/impl/CollectionDaoImpl.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.stereotype.Service;
+
+import edu.asu.diging.quadriga.core.mongo.CollectionDao;
+import edu.asu.diging.quadriga.domain.elements.Collection;
+
+@Service
+public class CollectionDaoImpl implements CollectionDao {
+
+    @Autowired
+    private MongoTemplate mongoTemplate;
```
jdamerow: Let's use Mongo Repositories for the typical CRUD actions.

quadriga/src/main/java/edu/asu/diging/quadriga/core/mongo/impl/CollectionDaoImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.stereotype.Service;
+
+import edu.asu.diging.quadriga.core.mongo.CollectionDao;
+import edu.asu.diging.quadriga.domain.elements.Collection;
+
+@Service
+public class CollectionDaoImpl implements CollectionDao {
```
jdamerow: Let's call this class `CollectionServiceImpl` (and adjust the interface accordingly)

quadriga/src/main/java/edu/asu/diging/quadriga/core/mongo/impl/CollectionDaoImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Autowired
+    private MongoTemplate mongoTemplate;
+
+    @Override
+    public void saveCollection(Collection collection) {
```
jdamerow: save method should return the saved collection

quadriga/src/main/java/edu/asu/diging/quadriga/service/impl/CollectionManager.java
```diff
@@ -0,0 +1,24 @@
...
+import edu.asu.diging.quadriga.core.mongo.CollectionDao;
+import edu.asu.diging.quadriga.domain.elements.Collection;
+import edu.asu.diging.quadriga.service.ICollectionManager;
+import edu.asu.diging.quadriga.web.forms.CollectionForm;
+
+@Service
+public class CollectionManager implements ICollectionManager {
+
+    @Autowired
+    private CollectionDao collectionDao;
```
jdamerow: what does the dao do other than the mongo template? unless there is additional things the dao class (service class) does, call the mongo repository from the manager.

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddCollectionController.java
```diff
@@ -0,0 +1,37 @@
...
+    @Autowired
+    private ICollectionManager collectionManager;
+
+    @RequestMapping(value = "/admin/collections/add", method = RequestMethod.GET)
+    public String get(Model model) {
+        model.addAttribute("collection", new CollectionForm());
+        return "admin/addCollection";
+    }
+
+    @RequestMapping(value = "/admin/collections/add", method = RequestMethod.POST)
```
jdamerow: let's put this part at `/auth` instead of `/admin` and users should have access to this part of the application.

jdamerow (changes requested):
tests and javadoc are missing

quadriga/src/main/java/edu/asu/diging/quadriga/service/impl/CollectionManager.java
```diff
@@ -0,0 +1,35 @@
...
+     * 
+     * 
+     * @return Collection Instance that is saved in database
+     * 
+     **/
+    public Collection addCollection(CollectionForm collectionForm) {
+        Collection collection = new Collection();
+        collection.setName(collectionForm.getName());
+        collection.setDescription(collectionForm.getDescription());
+        collectionRepo.save(collection);
```
jdamerow, line 31: you want to return the collection that the save method returns (I believe it is not void, right?) in case the caller needs the id of the newly created object (which is not guaranteed to be set in the passed object).

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddCollectionController.java
```diff
@@ -0,0 +1,37 @@
...
+    private ICollectionManager collectionManager;
+
+    @RequestMapping(value = "/auth/collections/add", method = RequestMethod.GET)
+    public String get(Model model) {
+        model.addAttribute("collection", new CollectionForm());
+        return "admin/user/addCollection";
+    }
+
+    @RequestMapping(value = "/auth/collections/add", method = RequestMethod.POST)
+    public String add(@ModelAttribute CollectionForm collectionForm, RedirectAttributes redirectAttrs) {
```
jdamerow: a collection should at least have a name; add a validator here.

jdamerow (changes requested):

jdamerow (changes requested):
javadoc is missing

jdamerow:
Make it so, Jenkins.

jdamerow:
There are test failures.
```
test_addCollection_success(edu.asu.diging.quadriga.core.service.impl.CollectionManagerTest)  Time elapsed: 1.513 sec  <<< FAILURE!
Argument(s) are different! Wanted:
collectionRepo.save(
    null
);
-> at edu.asu.diging.quadriga.core.service.impl.CollectionManagerTest.test_addCollection_success(CollectionManagerTest.java:44)
Actual invocations have different arguments:
collectionRepo.save(
    edu.asu.diging.quadriga.domain.elements.Collection@3f6b0be5
);
-> at edu.asu.diging.quadriga.service.impl.CollectionManager.addCollection(CollectionManager.java:31)

	at edu.asu.diging.quadriga.core.service.impl.CollectionManagerTest.test_addCollection_success(CollectionManagerTest.java:44)
```

quadriga/src/main/java/edu/asu/diging/quadriga/service/impl/CollectionManager.java
jdamerow: not addressed

jdamerow (changes requested):

PR #11, opened 2021-08-31
[story/Q20-11] completed edit collections functionality, added unit tests

Pooja444:

Ticket name: Users should be able to edit collections.
This ticket expects the implementation of full-stack changes to enable users of the Quadriga2.0 system to edit the collections they created.
  
* I have added a "TestLiterals" class for unit tests as it is always a good idea to have String constants (reduces build time, can be changed at one place to reflect throughout the project, etc.)
* I am not sure if Quadriga already does this, although I couldn't find it anywhere.
* I have added test cases for the new controller as well as for new methods in CollectionManager service
As per the discussion on slack with Julia,  I have fixed a bug that didn't show an appropriate alert after adding a new collection

### DIFF
--- a/quadriga/src/main/java/edu/asu/diging/quadriga/core/data/CollectionRepository.java
+++ b/quadriga/src/main/java/edu/asu/diging/quadriga/core/data/CollectionRepository.java
@@ -1,5 +1,6 @@
 package edu.asu.diging.quadriga.core.data;
 
+import org.bson.types.ObjectId;
 import org.springframework.data.mongodb.repository.MongoRepository;
 import org.springframework.stereotype.Repository;
 
@@ -7,5 +8,7 @@ import edu.asu.diging.quadriga.core.model.Collection;
 
 @Repository
 public interface CollectionRepository extends MongoRepository<Collection, String>{
+
+    Collection findBy_id(ObjectId _id);
 
 }
--- a/quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
+++ b/quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
@@ -1,5 +1,8 @@
 package edu.asu.diging.quadriga.core.service.impl;
 
+import java.util.Optional;
+
+import org.bson.types.ObjectId;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
@@ -13,7 +16,6 @@ public class CollectionManagerImpl implements CollectionManager {
     @Autowired
     private CollectionRepository collectionRepo;
 
-    
     /**
      * Creates a new Collection instance and stores it in mongodb
      * 
@@ -28,6 +30,44 @@ public class CollectionManagerImpl implements CollectionManager {
         collection.setName(name);
         collection.setDescription(description);
         return collectionRepo.save(collection);
+    }
+
+    /**
+     * Finds a collection from mongodb collection table by _id
+     * 
+     * @param id used to look up the collection in mongodb
+     * 
+     * 
+     * @return Collection Instance that is found from the database
+     * 
+     **/
+    @Override
+    public Collection findCollection(String id) {
+        return collectionRepo.findBy_id(new ObjectId(id));
+    }
+
+    /**
+     * Edits an existing Collection and updates it in mongodb
+     * 
+     * @param _id of the collection that needs to be updated
+     * 
+     * 
+     * @return Collection Instance that is updated in database
+     * 
+     **/
+    @Override
+    public Collection editCollection(String id, Optional<String> name, Optional<String> description) {
+        Optional<Collection> collection = Optional.ofNullable(findCollection(id));
+        Collection updatedCollection = null;
+
+        if (collection.isPresent()) {
+            Collection currentCollection = collection.get();
+            name.ifPresent(updatedName -> currentCollection.setName(updatedName));
+            description.ifPresent(updatedDescription -> currentCollection.setDescription(updatedDescription));
+            updatedCollection = collectionRepo.save(currentCollection);
+        }
+
+        return updatedCollection;
     }
 
 }
--- a/quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
+++ b/quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
@@ -1,0 +1,101 @@
+package edu.asu.diging.quadriga.web;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import javax.validation.Valid;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
+
+import edu.asu.diging.quadriga.core.model.Collection;
+import edu.asu.diging.quadriga.core.service.CollectionManager;
+import edu.asu.diging.quadriga.web.forms.CollectionForm;
+
+/**
+ * This controller enables users to edit their collections
+ * 
+ * @author poojakulkarni
+ *
+ */
+@Controller
+public class EditCollectionController {
+
+    @Autowired
+    private CollectionManager collectionManager;
+
+    /**
+     * Request handler for getting the "Edit collections" view
+     * 
+     * @param id    of the collection to be edited
+     * @param model to set collectionForm attribute to pre-fill current name and
+     *              description
+     * @return the editCollection view
+     */
+    @RequestMapping(value = "/auth/collections/edit", method = RequestMethod.GET)
+    public String get(@RequestParam(value = "id", required = true) String id, Model model) {
+        Optional<Collection> collection = Optional.ofNullable(collectionManager.findCollection(id));
+
+        if (collection.isPresent()) {
+            Collection currentCollection = collection.get();
+            CollectionForm collectionForm = new CollectionForm();
+            collectionForm.setId(id);
+            collectionForm.setName(currentCollection.getName());
+            collectionForm.setDescription(currentCollection.getDescription());
+            model.addAttribute("collectionForm", collectionForm);
+
+            return "auth/editCollection";
+        }
+
+        return "redirect:/auth/collections";
+    }
+
+    /**
+     * Request handler for editing a collection in the "Edit collections" view
+     * 
+     * @param id                 of the collection to be edited
+     * @param collectionForm     that has the updated name and description values
+     * @param bindingResult      for checking form errors
+     * @param redirectAttributes used for adding flash attributes after redirecting
+     * @return
+     */
+    @RequestMapping(value = "/auth/collections/edit", method = RequestMethod.POST)
+    public String edit(@RequestParam(value = "id", required = true) String id, @Valid CollectionForm collectionForm,
+            BindingResult bindingResult, RedirectAttributes redirectAttributes) {
+        if (bindingResult.hasErrors()) {
+            redirectAttributes.addFlashAttribute("alert_type", "danger");
+            redirectAttributes.addFlashAttribute("alert_msg",
+                    "Something went wrong while editing collections, please try again!");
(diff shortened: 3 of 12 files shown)


quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -30,4 +32,42 @@ public Collection addCollection(String name, String description) {
...
+     * 
+     **/
+    @Override
+    public Collection editCollection(String id, Optional<String> name, Optional<String> description) {
+        Optional<Collection> collection = Optional.ofNullable(findCollection(id));
+        Collection updatedCollection = null;
+
+        if (collection.isPresent()) {
+            Collection currentCollection = collection.get();
+            name.ifPresent(updatedName -> currentCollection.setName(updatedName));
```
Pooja444, line 65: FYI, I have added an Optional name and description, in case name and/or description are somehow passed as null, the app would still continue without a NullPointerException

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/CollectionRepository.java
```diff
@@ -8,4 +9,6 @@
 @Repository
 public interface CollectionRepository extends MongoRepository<Collection, String>{
 
+    Collection findBy_id(ObjectId _id);
```
jdamerow, line 12: I would change this repo to `MongoRepository<Collection, ObjectId>` and annotate `id` in Collection with `@Id`. You shouldn't need this method then.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -30,4 +32,42 @@ public Collection addCollection(String name, String description) {
...
+     * 
+     * @param _id of the collection that needs to be updated
+     * 
+     * 
+     * @return Collection Instance that is updated in database
+     * 
+     **/
+    @Override
+    public Collection editCollection(String id, Optional<String> name, Optional<String> description) {
+        Optional<Collection> collection = Optional.ofNullable(findCollection(id));
```
jdamerow, line 60: why do you put the returned into an optional here instead of just checking for null in line 63?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
jdamerow: setting name to null shouldn't throw an exception (there is no operation done on `name` if you just call `setName(name)` and in some cases it might be desired.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -30,4 +32,42 @@ public Collection addCollection(String name, String description) {
...
+        Collection updatedCollection = null;
+
+        if (collection.isPresent()) {
+            Collection currentCollection = collection.get();
+            name.ifPresent(updatedName -> currentCollection.setName(updatedName));
+            description.ifPresent(updatedDescription -> currentCollection.setDescription(updatedDescription));
```
jdamerow, line 66: same here

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -30,4 +32,42 @@ public Collection addCollection(String name, String description) {
...
+            updatedCollection = collectionRepo.save(currentCollection);
```
jdamerow, line 67: just return here and return null at the end of the method. In fact, I would throw an exception if a collection with the provided id doesn't exist.

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+     * Request handler for getting the "Edit collections" view
+     * 
+     * @param id    of the collection to be edited
+     * @param model to set collectionForm attribute to pre-fill current name and
+     *              description
+     * @return the editCollection view
+     */
+    @RequestMapping(value = "/auth/collections/edit", method = RequestMethod.GET)
+    public String get(@RequestParam(value = "id", required = true) String id, Model model) {
+        Optional<Collection> collection = Optional.ofNullable(collectionManager.findCollection(id));
```
jdamerow, line 43: same here, why not check agains null?

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+            CollectionForm collectionForm = new CollectionForm();
+            collectionForm.setId(id);
+            collectionForm.setName(currentCollection.getName());
+            collectionForm.setDescription(currentCollection.getDescription());
+            model.addAttribute("collectionForm", collectionForm);
+
+            return "auth/editCollection";
+        }
+
+        return "redirect:/auth/collections";
```
jdamerow, line 56: there should be a 404 not found page shown to the user if the collection doesn't exist (not the tomcat one but a custom quadriga 404 one).

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    @RequestMapping(value = "/auth/collections/edit", method = RequestMethod.POST)
+    public String edit(@RequestParam(value = "id", required = true) String id, @Valid CollectionForm collectionForm,
+            BindingResult bindingResult, RedirectAttributes redirectAttributes) {
+        if (bindingResult.hasErrors()) {
+            redirectAttributes.addFlashAttribute("alert_type", "danger");
+            redirectAttributes.addFlashAttribute("alert_msg",
+                    "Something went wrong while editing collections, please try again!");
+            redirectAttributes.addFlashAttribute("show_alert", true);
+            return "auth/editCollection";
```
jdamerow, line 76: the form just needs to show the validation results for the fields with the previously entered values still present.

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        Collection collection = null;
+
+        if (Objects.nonNull(collectionForm)) {
```
jdamerow, line 81: I don't think collectionForm can ever be null.

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -0,0 +1,101 @@
...
+            collection = collectionManager.editCollection(id, Optional.ofNullable(collectionForm.getName()),
```
jdamerow, line 82: merge with line 79

quadriga/src/test/java/edu/asu/diging/quadriga/core/literals/TestLiterals.java
```diff
@@ -0,0 +1,18 @@
+package edu.asu.diging.quadriga.core.literals;
+
+public class TestLiterals {
+    public static final String BLANK = "";
+
+    public static final class CollectionLiterals {
+        public static final String COLLECTION_NAME = "Collection name";
+        public static final String COLLECTION_DESC = "Collection description";
+        public static final String EDITED_NAME = "Edited name";
+        public static final String EDITED_DESC = "Edited description";
```
jdamerow, line 10: These should just be constants in the test class since we won't need them in any other test classes most likely.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -44,11 +48,215 @@ public boolean matches(Collection arg0) {
...
+        collection.setId(objectId);
+        Mockito.when(collectionRepo.findBy_id(objectId)).thenReturn(collection);
+        Collection foundCollection = managerToTest.findCollection(objectId.toString());
+        Assert.assertEquals(collection.getId().toString(), foundCollection.getId().toString());
+    }
+
+    @Test
+    public void test_editCollection_success() {
+
+        String name = CollectionLiterals.COLLECTION_NAME;
```
jdamerow, line 71: use the constant where it is needed.

jdamerow: same everywhere else, there is no point in creating a new variable for it if it's only used once afterwards.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -44,11 +48,215 @@ public boolean matches(Collection arg0) {
...
+        String desc = CollectionLiterals.COLLECTION_DESC;
+        Collection existingCollection = new Collection();
+        ObjectId id = new ObjectId();
+        existingCollection.setId(id);
+        existingCollection.setName(name);
+        existingCollection.setDescription(desc);
+
+        Mockito.when(managerToTest.findCollection(id.toString())).thenReturn(existingCollection);
```
jdamerow, line 79: managerToTest is the class you're trying to test. Don't mock it.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -44,11 +48,215 @@ public boolean matches(Collection arg0) {
...
+
+        Optional<String> editedName = Optional.of(CollectionLiterals.EDITED_NAME);
+        Optional<String> editedDescription = Optional.of(CollectionLiterals.EDITED_DESC);
```
jdamerow, line 82: why are you creating optionals here?

jdamerow (changes requested):

jdamerow:
also, please don't force push.  There shouldn't be any need for that.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
Pooja444: Okay, got it. Will remove Optional

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
Pooja444: I was creating them. since the method accepts optional. But now that I'll be removing Optional, I'll change this too.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
Pooja444: I had added it as I am using it in two classes: EditCollectionControllerTest and CollectionManagerImplTest
Should I add these constants separately in these two classes?

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
Pooja444: Oh alright, I should be mocking the repo, got it.
Will update this

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
Pooja444: Should I create one or does one exist already?

jdamerow: You have to check. There probably is not yet one.

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
Pooja444: Didn't get you here?

The form does display pre-filled values while editing a collection and an alert in case the form is invalid.
The flash attributes here are added for displaying the result (success/failure) on UI after the editing is complete (i.e. after redirection to Show Collections page)

PR #13, opened 2021-09-07
Story/q20 12

Pooja444:

When the network is being submitted, there needs to be specified to which collection (probably as part of the url). then the network would be linked to that collection (via properties in neo4j and mongodb).
  
As this change is only made in one of the controllers, I've not added any unit test cases

### DIFF
--- a/quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
+++ b/quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
@@ -56,7 +56,7 @@ public class AddNetworkApiController {
         Collection collection = collectionManager.findCollection(collectionId);
         
         if(Objects.isNull(collection)) {
-            return HttpStatus.BAD_REQUEST;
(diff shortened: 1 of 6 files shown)


quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -44,7 +51,13 @@
      */
     @ResponseBody
     @RequestMapping(value = "/api/v1/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
-    public HttpStatus processJson(@RequestBody Quadruple quadruple) {
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @RequestParam(value = "collectionid", required = true) String collectionId) {
```
jdamerow, line 54: Let's make this more REST-like and make the collection id part of the path instead of a request parameter.

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

PR #16, opened 2021-09-07
Story/q20 10

Pooja444:

Users should be able to delete collections
  
Added a modal that confirms deletion action before actually deleting a collection


Pooja444:
Force pushed my changes as I did a "pull --rebase" from story/Q20-11 branch to resolve conflicts before doing my latest changes. Git won't let me do a normal push after rebasing, so had to force push

jdamerow:
Please don't rebase a branch that has already been pushed (if someone else is working with the same branch you can really mess up things fro them) . Then you also shouldn't need to force push.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -67,5 +67,21 @@ public Collection editCollection(String id, String name, String description) thr
...
+     * @param id used to look up the collection in mongodb
+     * 
+     */
+    public void deleteCollection(String id) throws CollectionNotFoundException {
+        Collection collection = findCollection(id);
+        
+        if(Objects.nonNull(collection)) {
+            collectionRepo.delete(collection);
+        } else {
+            throw new CollectionNotFoundException();
```
MaulikLim, line 87: In case if the collection does not exist, we would just return instead of throwing an exception. Could you verify @jdamerow .

MaulikLim (commented):
If the exception is removed, there would changes in controller and test cases.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
jdamerow: either approach is fine with me; you can either argue that deleting something that doesn't exist (aka is already deleted) should just behave like it's successfully deleted as the outcome is the same. Or you can say trying to delete something that doesn't exist should be made known. Either is fine with me as long as it's consistent across the application (meaning other types of objects should behave the same way).

MaulikLim: In that case, current version should be fine

MaulikLim (approved):

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -61,6 +61,48 @@ public boolean matches(Collection arg0) {
...
+        String desc = "description";
+        Collection collection = new Collection();
+        ObjectId id = new ObjectId();
+        collection.setId(id);
+        collection.setName(name);
+        collection.setDescription(desc);
+        
+        Mockito.when(collectionRepo.findById(id)).thenReturn(Optional.ofNullable(null));
+        
+        Exception e = Assert.assertThrows(CollectionNotFoundException.class,
```
jdamerow: Is there a reason you assign the exception to a variable?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -68,5 +68,21 @@ public Collection editCollection(String id, String name, String description) thr
...
+     * Deletes a collection from mongodb collection table by _id
+     * 
+     * @param id used to look up the collection in mongodb
+     * 
+     */
+    public void deleteCollection(String id) throws CollectionNotFoundException {
+        Collection collection = findCollection(id);
+        
+        if(Objects.nonNull(collection)) {
+            collectionRepo.delete(collection);
```
jdamerow, line 85: let's make a comment here that.  this will have to be adjusted once there are networks in the collection.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -68,5 +68,21 @@ public Collection editCollection(String id, String name, String description) thr
             throw new CollectionNotFoundException();
         }
     }
+    
+    /**
+     * Deletes a collection from mongodb collection table by _id
```
jdamerow: this method is not mongodb specific and shouldn't mention mongo db here. If we switch backend, this method should (in theory) still work the same way. (the repository is mongo db specific).

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

PR #17, opened 2021-09-07
Story/q20 12

Pooja444:

when the network is being submitted, there needs to be specified to which collection (probably as part of the url). then the network would be linked to that collection (via properties in neo4j and mongodb).
  
... describe here ...


quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -44,7 +51,13 @@
      */
     @ResponseBody
     @RequestMapping(value = "/api/v1/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
-    public HttpStatus processJson(@RequestBody Quadruple quadruple) {
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @RequestParam(value = "collectionid", required = true) String collectionId) {
+        
+        Collection collection = collectionManager.findCollection(collectionId);
+        
+        if(Objects.isNull(collection)) {
```
MaulikLim: Is there any specific reason to use `Objects.isNull()?` Wouldn't simple null check work as well?

MaulikLim (commented):
I am not sure if this intended and how it would work based on which is merged first, but you are merging into other branch which has a separate pull request.

Pooja444:
> I am not sure if this intended and how it would work based on which is merged first, but you are merging into other branch which has a separate pull request.

Yes that's because there are overlapping files between this and the PR for story Q20-11. Had discussed it with Julia previously.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
Pooja444: Yes that would work as well, should I update it?
I used to follow this practice earlier so probably did it out of habit :)

MaulikLim: I think simple check would make it uniform as it is more commonly used through out our apps

MaulikLim (approved):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -44,7 +50,13 @@
      */
     @ResponseBody
     @RequestMapping(value = "/api/v1/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
-    public HttpStatus processJson(@RequestBody Quadruple quadruple) {
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @RequestParam(value = "collectionid", required = true) String collectionId) {
```
jdamerow: let's make the collection id a path variable, e.g. `/api/v1/collection/{collectionId}/network/add`

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -44,7 +50,13 @@
...
+        
+        Collection collection = collectionManager.findCollection(collectionId);
+        
+        if(collection == null) {
+            return HttpStatus.NOT_ACCEPTABLE;
```
jdamerow: This status code is in cases of: "The requested resource is capable of generating only content not acceptable according to the Accept headers sent in the request."

Here it should either return a not found (if the collection is part of the url) or bad request (if the request is a request parameter)

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -43,8 +49,18 @@
      * @return
      */
     @ResponseBody
-    @RequestMapping(value = "/api/v1/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
-    public HttpStatus processJson(@RequestBody Quadruple quadruple) {
+    @RequestMapping(value = "/api/v1/collection/{collectionId}/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @PathVariable String collectionId) {
+        
+        if(collectionId == null || collectionId.trim().isEmpty()) {
```
jdamerow: Have you tried this? I'm fairly certain this couldn't happen as the url pattern wouldn't match this case, but I might be wrong.

jdamerow (changes requested):
- There seem to be files in here that do not belong to this ticket. It might just be a github issue and you just might have to create a new PR.
- This does not seem to be addressed for the mapped networks (neo4j): "then the network would be linked to that collection (via properties in neo4j and mongodb)."

jdamerow:
Resolve conflicts please

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
Pooja444: Yes this is not required, we just get NotFound because the pattern doesn't match
Will remove this

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -43,7 +43,11 @@ public Collection addCollection(String name, String description) {
      **/
     @Override
     public Collection findCollection(String id) {
-        return collectionRepo.findById(new ObjectId(id)).orElse(null);
+        try {
+            return collectionRepo.findById(new ObjectId(id)).orElse(null);
+        } catch(IllegalArgumentException e) {
```
jdamerow, line 50: don't swallow the exception, it should be logged.

jdamerow (changes requested):

jdamerow:
> This does not seem to be addressed for the mapped networks (neo4j): "then the network would be linked to that collection (via properties in neo4j and mongodb)."

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -43,8 +51,18 @@
...
+    @RequestMapping(value = "/api/v1/collection/{collectionId}/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @PathVariable String collectionId) {
+        
+        MappedCollection mappedCollection;
+        try {
+            mappedCollection = mappedCollectionService.findOrAddMappedCollectionByCollectionId(collectionId);
+            if(mappedCollection == null) {
+                return HttpStatus.NOT_FOUND;
+            }
+        } catch(InvalidObjectIdException | CollectionNotFoundException e)  {
```
jdamerow, line 72: exception needs to be logged

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedCollection.java
```diff
@@ -0,0 +1,46 @@
...
+import org.bson.types.ObjectId;
+
+/**
+ * A mapped collection is used to map Concepts and Predicates to a collection
+ * Once a network is submitted by a user, the user can give a name to the mapped collection
+ * 
+ * @author poojakulkarni
+ *
+ */
+public class MappedCollection {
```
jdamerow: There is already a collection class that should be used.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -39,11 +44,17 @@ public Collection addCollection(String name, String description) {
...
      * 
      **/
     @Override
-    public Collection findCollection(String id) {
-        return collectionRepo.findById(new ObjectId(id)).orElse(null);
+    public Collection findCollection(String id) throws InvalidObjectIdException {
+        try {
+            return collectionRepo.findById(new ObjectId(id)).orElse(null);
+        } catch(IllegalArgumentException e) {
+            logger.error("Couldn't retrieve a collection, invalid collectionId: " + id);
```
jdamerow: the whole exception needs to be logged; you're loosing the stack trace here.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -55,16 +66,18 @@ public Collection findCollection(String id) {
...
-    public Collection editCollection(String id, String name, String description) throws CollectionNotFoundException {
+    public Collection editCollection(String id, String name, String description) throws CollectionNotFoundException, InvalidObjectIdException {
         Collection collection = findCollection(id);
 
         if (Objects.nonNull(collection)) {
             collection.setName(name);
             collection.setDescription(description);
             return collectionRepo.save(collection);
         } else {
+            logger.error("Couldn't find collection for collectionId: " + id);
```
jdamerow: this is not necessary since an exception is thrown.

jdamerow: the id of the collection should be passed into the exception though

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -84,6 +98,7 @@ public void deleteCollection(String id) throws CollectionNotFoundException {
             // If it is linked to a network, we will archive the collection.
             collectionRepo.delete(collection);
         } else {
+            logger.error("Couldn't find collection for collectionId: " + id);
```
jdamerow: see above

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedCollectionServiceImpl.java
```diff
@@ -0,0 +1,164 @@
...
+     *                                     ObjectId
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     does't exist
+     */
+    private Collection checkAndGetCollection(String collectionId)
+            throws InvalidObjectIdException, CollectionNotFoundException {
+        Collection collection = collectionManager.findCollection(collectionId);
+        if (collection == null) {
+            logger.error("Couldn't find collection for collectionId: " + collectionId);
+            throw new CollectionNotFoundException();
```
jdamerow: same as above, no error message but pass id into exception

jdamerow (changes requested):
I'll stop with the review as I think there is some confusion here. The ticket says: 
> when the network is being submitted, there needs to be specified to which collection (probably as part of the url). then the network would be linked to that collection (via properties in neo4j and mongodb).

The collection class already exists, for this ticket new graphs (and hence mapped triples) should be added to these collections. Mapped Triples will in addition be grouped by there mappings (let's call these Mapping Groups). For now there are only the default mappings, so all triples would belong to a "Default Mapping Group", but later there could be other mapping groups.
Does this make sense?

jdamerow:
Thinking about this again, i think it's close just a little confusing:
- I would call the mappedCollections mappingGroups or mappedTripleGroups or something to not confuse it with collections
- it needs to be possible to mark a mapping group as the default mapping group and that's the one to which the default mapped triples are being added to in the add network controller

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedCollection.java
Pooja444: Hi Julia,

When I started the ticket and created "MappedCollection" as per this comment:

Julia Lessios-Damerow  11:44 AM
There needs to be a connection between the collections and the mapped networks (and the mappings for that matter). I think I would introduce a new MappedCollection or something like that belongs to a collection and then all the concepts belong to a mapped collection. something like that. Does that make sense?

Should I remove it now?

Pooja444:
@jdamerow Got it, I will make these changes
I will rename MappedCollections to MappedTripleGroups as it is more clearer and keep a enum field to mark the mappedTriple as Default

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -43,8 +56,21 @@
...
+        // Every time a new network is submitted, the triple in that network has to be added as a
+        // new MappedTripleGroup for given collectionId
+        MappedTripleGroup mappedTripleGroup;
+        try {
+            mappedTripleGroup = mappedTripleGroupService.addMappedTripleGroup(collectionId);
+            if(mappedTripleGroup == null) {
+                return HttpStatus.NOT_FOUND;
+            }
+        } catch(InvalidObjectIdException | CollectionNotFoundException e)  {
+            logger.error(ExceptionUtils.getStackTrace(e));
```
jdamerow: just pass the exception into `error`

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -43,8 +56,21 @@
      * @return
      */
     @ResponseBody
-    @RequestMapping(value = "/api/v1/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
-    public HttpStatus processJson(@RequestBody Quadruple quadruple) {
+    @RequestMapping(value = "/api/v1/collection/{collectionId}/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
+    public HttpStatus processJson(@RequestBody Quadruple quadruple, @PathVariable String collectionId) {
+        
+        // Every time a new network is submitted, the triple in that network has to be added as a
+        // new MappedTripleGroup for given collectionId
```
jdamerow: that is incorrect. All networks have to be added to the default mapped triple group.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -53,11 +79,15 @@ public HttpStatus processJson(@RequestBody Quadruple quadruple) {
         // save network
         List<CreationEvent> events = networkMapper.mapNetworkToEvents(quadruple.getGraph());
         List<EventGraph> eventGraphs = events.stream().map(e -> new EventGraph(e)).collect(Collectors.toList());
-        eventGraphs.forEach(e -> e.setDefaultMapping(quadruple.getGraph().getMetadata().getDefaultMapping()));
+        eventGraphs.forEach(e -> {
+            e.setCollectionId(mappedTripleGroup.getCollectionId());
```
jdamerow: why don't you use `collectionId` here? This makes it look like `collectionId` is different than `mappedTripleGroup.getCollectionId()`.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedTripleGroup.java
```diff
@@ -0,0 +1,71 @@
+package edu.asu.diging.quadriga.core.model;
+
+import javax.persistence.Id;
+
+import org.bson.types.ObjectId;
+
+/**
+ * A MappedTripleGroup is used to map Concepts and Predicates to a collection
```
jdamerow: that's incorrect. A mapped triple group groups together all the mapped triples that belong to specific mappings (e.g. all the triples resulting from default mappings, or all the triples that results from a user specified set of mappings).

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedTripleGroup.java
```diff
@@ -0,0 +1,71 @@
...
+ * One CollectionId can have multiple MappedTripleGroups
+ * A new MappedTripleGroup is created per network
```
jdamerow: no, a mapped triple group contains multiple triples.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedTripleGroup.java
```diff
@@ -0,0 +1,71 @@
...
+ * Every MappedTripleGroup will have MappedTripleType which specifies mapping type
+ * Once a network is submitted by a user, the user can give a name to the mapped collection
```
jdamerow: do you man "give a name to the mapped triple group"?

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/MappedTripleType.java
```diff
@@ -0,0 +1,5 @@
+package edu.asu.diging.quadriga.core.model;
+
+public enum MappedTripleType {
+    DefaultMapping
```
jdamerow: enum values are all uppercase and snake case

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -39,11 +40,16 @@ public Collection addCollection(String name, String description) {
...
      * 
      **/
     @Override
-    public Collection findCollection(String id) {
-        return collectionRepo.findById(new ObjectId(id)).orElse(null);
+    public Collection findCollection(String id) throws InvalidObjectIdException {
+        try {
+            return collectionRepo.findById(new ObjectId(id)).orElse(null);
+        } catch(IllegalArgumentException e) {
+            throw new InvalidObjectIdException("CollectionId: " + id);
```
jdamerow: you're still swallowing the exception. The stack trace will be gone.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+     * collectionId shouldn't be linked to the mappedTripleGroup
+     * 
+     * @param collectionId is the id of the collection to be checked
+     * @return the Collection entry found in the database
+     * @throws InvalidObjectIdException    if collectionId couldn't be conveted to
+     *                                     ObjectId
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     does't exist
+     */
+    private Collection checkAndGetCollection(String collectionId)
```
jdamerow: why is this not just called `get` or `find` or `getCollection`?

jdamerow: wait, and why is this method here? Why not use the collection manager?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public MappedTripleGroup addMappedTripleGroup(String collectionId)
+            throws InvalidObjectIdException, CollectionNotFoundException {
+        
+        MappedTripleGroup mappedTripleGroup = new MappedTripleGroup();
+        
+        mappedTripleGroup.setCollectionId(checkAndGetCollection(collectionId).getId());
+        
+        // This will be updated when custom mappings are added
+        mappedTripleGroup.setMappedTripleType(MappedTripleType.DefaultMapping);
```
jdamerow: type should be passed into the method, so this method doesn't need to be changed for custom triple groups.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+     * 
+     * @param collectionId used to look for the MappedTripleGroup entry
+     * @return the MappedTripleGroup entry that was found, else null
+     * @throws InvalidObjectIdException    if collectionId couldn't be converted to
+     *                                     ObjectId
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     doesn't exist
+     */
+    @Override
+    public MappedTripleGroup findMappedTripleGroupByCollectionId(String collectionId)
```
jdamerow: there could be multiple mapped triple groups for a collection

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+     * This method tries to find a MappedTripleGroup with the given
+     * mappedTripleGroupId
+     * 
+     * @param mappedTripleGroupId is the id used to find a MappedTripleGroup
+     * @return the found mappedTripleGroupEntry
+     * @throws InvalidObjectIdException if mappedTripleGroupId couldn't be converted
+     *                                  to ObjectId
+     */
+    @Override
+    public MappedTripleGroup findMappedTripleGroupById(String mappedTripleGroupId) throws InvalidObjectIdException {
```
jdamerow: just call this `get` or `getById` or something

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: Hi Julia,
I added this method here because it is supposed to throw an exception if a collection is not found because it means that a network is being submitted to a collection that doesn't exist
CollectionManager already has a findCollection method that simply finds and returns the collection and doesn't throw an exception
So I can't rename it to findCollection, probably I'll rename it to getCollection and add it to the CollectionManager, but I'm afraid that getCollection and findCollection, can be confusing

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
Pooja444: Since I can't directly do `logger.error(e)` as the type of 'e' is 'Exception', I have to pass e.getMessage() to `error`

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -43,8 +57,21 @@
...
+        // Every time a new network is submitted, the triple in that network has to be added as the
+        // default MappedTripleGroup for given collectionId
+        MappedTripleGroup mappedTripleGroup;
+        try {
+            mappedTripleGroup = mappedTripleGroupService.getMappedTripleGroup(collectionId, MappedTripleType.DEFAULT_MAPPING);
+            if(mappedTripleGroup == null) {
+                return HttpStatus.NOT_FOUND;
+            }
+        } catch(InvalidObjectIdException | CollectionNotFoundException e)  {
+            logger.error(e.getMessage());
```
jdamerow: you're swallowing the stacktrace, which makes it hard to impossible to figure out what went wrong. You have to pass the exception.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+     */
+    @Override
+    public MappedTripleGroup addMappedTripleGroup(String collectionId, MappedTripleType mappedTripleType)
+            throws InvalidObjectIdException, CollectionNotFoundException {
+        
+        MappedTripleGroup mappedTripleGroup = new MappedTripleGroup();
+        
+        mappedTripleGroup.setCollectionId(collectionManager.getCollection(collectionId).getId());
+        
+        // This will be updated when custom mappings are added
```
jdamerow: what will be updated?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+     * @throws InvalidObjectIdException if mappedTripleGroupId couldn't be converted
+     *                                  to ObjectId
+     */
+    @Override
+    public MappedTripleGroup getById(String mappedTripleGroupId) throws InvalidObjectIdException {
+        ObjectId mappedTripleGroupObjectId;
+        try {
+            mappedTripleGroupObjectId = new ObjectId(mappedTripleGroupId);
+        } catch (IllegalArgumentException e) {
+            throw new InvalidObjectIdException("MappedTripleGroupId: " + mappedTripleGroupId);
```
jdamerow: you need to keep the stack trace; pass the exception.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+import edu.asu.diging.quadriga.core.exceptions.CollectionNotFoundException;
+import edu.asu.diging.quadriga.core.exceptions.InvalidObjectIdException;
+import edu.asu.diging.quadriga.core.exceptions.MappedTripleGroupNotFoundException;
+import edu.asu.diging.quadriga.core.model.MappedTripleGroup;
+import edu.asu.diging.quadriga.core.model.MappedTripleType;
+import edu.asu.diging.quadriga.core.service.CollectionManager;
+import edu.asu.diging.quadriga.core.service.MappedTripleGroupService;
+
+@Service
+public class MappedTripleGroupServiceImpl implements MappedTripleGroupService {
```
jdamerow, line 18: since all the methods in this class are about mapped triple groups, the method names do not need to contain that. E.g. `updateMappedTripleGroupNameById` => `updateName`

quadriga/src/main/java/edu/asu/diging/quadriga/web/DeleteCollectionController.java
```diff
@@ -25,7 +26,7 @@ public String get(@PathVariable String id, RedirectAttributes redirectAttributes
             redirectAttributes.addFlashAttribute("show_alert", true);
         
             return "redirect:/auth/collections";
-        } catch (CollectionNotFoundException e) {
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
```
jdamerow, line 33: exception should be logged

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -40,17 +41,20 @@
...
+                CollectionForm collectionForm = new CollectionForm();
+                collectionForm.setId(id);
+                collectionForm.setName(collection.getName());
+                collectionForm.setDescription(collection.getDescription());
+                model.addAttribute("collectionForm", collectionForm);
+                return "auth/editCollection";
+            } else {
+                return "error404Page";
+            }
+        } catch (InvalidObjectIdException e) {
```
jdamerow, line 61: exception should be logged

jdamerow (changes requested):
-  for all javadoc, sentences end with a period. ;)

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
Pooja444: The logger.error() method won't accept an Object of type Exception, I tried
So I had added ExceptionUtils.getStackTrace which helps maintain stack trace but in one of the PR comments you asked me to remove that
So I added e.getMessage()
I am not sure what exactly to do here
Is there any example I could check?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
jdamerow: I believe `error` takes a string and then. an exception.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
jdamerow: I'm getting very  confused with t he different PRs.  Here is what I said on the other one:
why don't you use the collection manager and then check for null instead of creating its own method? This seems unnecessarily complicated.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -84,8 +93,30 @@ public void deleteCollection(String id) throws CollectionNotFoundException {
...
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     does't exist
+     */
+    @Override
+    public Collection getCollection(String collectionId) throws InvalidObjectIdException, CollectionNotFoundException {
+        Collection collection = findCollection(collectionId);
+        if (collection == null) {
+            throw new CollectionNotFoundException("CollectionId: " + collectionId);
         }
+        return collection;
```
jdamerow: this method should just return null if the collection doesn't exist. Retrieving an object that doesn't exist is something that could easily be used across the application (simply to check if a collection exist) and exceptions shouldn't be used in such cases (too expensive to keep building a stacktrace and more annoying to incorporate in the code).

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     doesn't exist
+     */
+    @Override
+    public MappedTripleGroup findByCollectionIdAndMappingType(String collectionId, MappedTripleType mappedTripleType)
+            throws InvalidObjectIdException, CollectionNotFoundException {
+        return mappedTripleGroupRepository
+                .findByCollectionIdAndMappedTripleType(collectionManager
+                        .getCollection(collectionId)
+                        .getId(), mappedTripleType)
```
jdamerow: Why are you retrieving the collection here? Why can't you just create an objectId out of the colletionId directly?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        MappedTripleGroup mappedTripleGroup = findByCollectionIdAndMappingType(collectionId, mappedTripleType);
+
+        // In case this is a new collection, or existing collection but new mapping type for that collection
+        // we create a new MappedTripleGroup entry
+        if (mappedTripleGroup == null) {
+            mappedTripleGroup = add(collectionId, mappedTripleType);
+        }
+        if (mappedTripleGroup == null) {
+            logger.error("Couldn't find or persist a new MappedTripleGroup entry for collectionId: " + collectionId);
+        }
```
jdamerow: this if is odd here. It just adds a log message but proceeds. That error message should be logged where the method is called an its result is handled.

jdamerow: In fact, when does this situation occur?

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
Pooja444: After a network is submitted for a collection, two scenarios could occur -
1. The collection exists, and it's the first time a network is being submitted to it
2. The collection doesn't exist, i.e. the network submission request is incorrect

In the first case, we add new mappedTripleGroup to the existing collection
In the second case, we return null, which will send a NOT_FOUND response to the request

Now here, if we don't retrieve the collection and simply pass `new ObjectId(collectionId)`, this method will return null for both the above cases
To distinguish between the two, we need to retrieve the collection here

I can change this method to -
Try to get Collection, and if it doesn't exist, throw CollectionNotFoundException here
If it exists, null will be returned, which means we need to add new MappedTripleGroup

Is that fine?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: After a network is submitted for a collection, two scenarios could occur -

1. The collection exists, and it's the first time a network is being submitted to it
2. The collection doesn't exist, i.e. the network submission request is incorrect

In the first case, we add new mappedTripleGroup to the existing collection
In the second case, we return null, which will send a NOT_FOUND response to the request

Now here, if we don't retrieve the collection and simply pass `new ObjectId(collectionId)`, this method will return null for both the above cases
To distinguish between the two, we need to retrieve the collection here

I can change this method to -
Try to get Collection, and if it doesn't exist, throw CollectionNotFoundException here
If it exists, null will be returned, which means we need to add new MappedTripleGroup

Is that fine?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: This situation could occur if something fails while inserting a new entry for mappedTripleGroup
But I can remove this altogether and just add the null check in the network submission request handler
Should I do that?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: Something like this -
```java
@Override
public MappedTripleGroup findByCollectionIdAndMappingType(String collectionId, MappedTripleType mappedTripleType)
    throws InvalidObjectIdException, CollectionNotFoundException {
    Collection collection = collectionManager.findCollection(collectionId);
        
    if(collection == null) {
        throw new CollectionNotFoundException("Couldn't find collection for id: " + collectionId);
    }
        
    return mappedTripleGroupRepository
            .findByCollectionIdAndMappedTripleType(collection.getId(), mappedTripleType)
            .orElse(null);
}
```

jdamerow: I"m confused. This method is called `findByCollectionIdAndMappingType` which indicates it simply returns a collection with the given id and mapping type. I expect this method to not return anything if the id and mapping type doesn't exist. What the caller does with that info is a different story. They can decide to create a new collection or raise a not found. I would make it the job of the caller to ensure the collection exists, but if you want to keep the check here and then, the above solution is better.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
jdamerow: I would expect that there would be an exception in that case?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: Yes, this method does return `null` if the collectionId + mapping type doesn't exist
The part where new MappedTripleEntry is created happens in the calling method, not here
I'll keep this logic and re-open the PR, let me know if that looks correct

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedTripleGroupServiceImpl.java
Pooja444: There's no exception raised by SimpleMongoRepository if saving fails
So, I have just added null check in the `AddNetworkApiController` itself and added logging there

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
Pooja444: Yes that makes sense. In that case I'll just remove this method altogether and just utilize the existing `findCollection` method

jdamerow:
Make it so, Jenkins.

jdamerow:
I can't really check if the networks are stored correctly since the collection don't show content yet, but it looks good. I'lll merge it and if something doesn't work we'll need to fix it later. Also, we'll need to do something about the collection ids, but I'll create a separate ticket for that.

PR #19, opened 2021-09-30
Story/q20 12

Pooja444:

when the network is being submitted, there needs to be specified to which collection (probably as part of the url). then the network would be linked to that collection (via properties in neo4j and mongodb).
  
In CollectionManager, I have also added an InvalidObjectIdException which will be thrown if CollectionId passed is not a valid string of hex characters. This affected EditCollectionController and DeleteCollectionController, so I have modified a few related files as well.


quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/MappedCollectionServiceImpl.java
```diff
@@ -0,0 +1,164 @@
...
+     * collectionId shouldn't be linked to the mappedCollection
+     * 
+     * @param collectionId is the id of the collection to be checked
+     * @return the Collection entry found in the database
+     * @throws InvalidObjectIdException    if collectionId couldn't be conveted to
+     *                                     ObjectId
+     * @throws CollectionNotFoundException if collection with given collectionId
+     *                                     does't exist
+     */
+    private Collection checkAndGetCollection(String collectionId)
```
jdamerow, line 41: There was a comment here before, right? why don't you use the collection manager and then check for null instead of creating its own method? This seems unnecessarily complicated.

Pooja444: This is actually an old PR that has not been updated since September 30. Instead of re-opening a PR for this story, I had created a new one and you asked me to close it, so this is the closed one

On the PR that is currently under review by you (https://github.com/diging/quadriga-2.0/pull/17), I have removed this method

PR #20, opened 2021-10-01
Story/q20 7

MaulikLim:

[Q20-7] - Admins should assign citesphere apps to collections
  
... describe here ...


MaulikLim:
Renamed `CollectionManager` to `ICollectionManager` and `CollectionManagerImpl` to `CollectionManager` and moved javadocs to the interface i.e., `ICollectionManager`.

quadriga/src/main/java/edu/asu/diging/quadriga/core/citesphere/impl/CitesphereConnector.java
```diff
@@ -0,0 +1,135 @@
...
+                return restTemplate.exchange(appUrl, HttpMethod.GET, entity,
+                        new ParameterizedTypeReference<List<CitesphereAppInfo>>() {
+                        }).getBody();
+            }
+            throw ex;
+        }
+    }
+
+    private String getAccessToken() {
+        AuthorizationGrant clientGrant = new ClientCredentialsGrant();
```
Pooja444: I have written the exact same method (branch story/Q20-3) in a 'TokenValidatorImpl'
Since you're writing that same method here, should I remove it from my changes and use this method instead?
In that case, you might have to make this method public

MaulikLim: I will make it public then.

MaulikLim: @Pooja444 The method should be available now

Pooja444 (approved):

quadriga/src/main/java/edu/asu/diging/quadriga/core/citesphere/impl/CitesphereConnector.java
jdamerow: @Pooja444, @MaulikLim depending on what the `TokenValidatorImpl` does the classes should probably be better encapsulated. Since `getCitesphereApps` is implemented here (using this method) and there is another method in a different class that also uses `getAccessToken` it sounds like you either want all methods connecting to citesphere (using `getAccessToken`) in this class or have low level methods like `getAccessToken` in it's own class. Does that make sense?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/ICollectionManager.java
```diff
@@ -0,0 +1,45 @@
+package edu.asu.diging.quadriga.core.service;
+
+import java.util.List;
+
+import edu.asu.diging.quadriga.core.exceptions.CollectionNotFoundException;
+import edu.asu.diging.quadriga.core.model.Collection;
+
+public interface ICollectionManager {
```
jdamerow: Is there a reason this has been renamed? It looks like the other service classes follow the convention interface does not start with `I` and implementation classes end with `Impl`.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/citesphere/impl/CitesphereConnector.java
MaulikLim: Yes it does. We'll discuss and make the changes.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/ICollectionManager.java
MaulikLim: Yeah. I got confused with the naming convention from the other project. I'll change it back.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -35,5 +38,13 @@ public String getDescription() {
...
         this.description = description;
     }
+    
+    public List<String> getApps() {
+        return apps;
+    }
+
+    public void setApps(List<String> apps) {
+        this.apps = apps;
+    }
```
jdamerow, line 52: the apps property needs some javadoc. is this the id of the app in citesphere? its name? something else?

jdamerow (changes requested):
Two thoughts on this:
- I think I would make the app part of adding/editing a collection asynchronous. Right now if citesphere is offline or just takes a long time, then the edit/add page would just throw an error.
- I think there should be some extra check before storing the apps with the collections to make sure the passed id (it's the id, right?) does actually exist and the app hasn't been deleted between retrieving the apps and storing a collection.

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddCollectionController.java
```diff
@@ -19,19 +25,34 @@
...
+        List<CitesphereAppInfo> citesphereApps = citesphereConnector.getCitesphereApps();
+        HashSet<String> appSet = new HashSet<String>(
+                citesphereApps.stream().map(app -> app.getClientId()).collect(Collectors.toList()));
+        
+        for (String clientId : collectionForm.getApps()) {
+            if (!appSet.contains(clientId)) {
+                model.addAttribute("collectionForm", collectionForm);
+                return "auth/addCollection";
+            }
+        }
```
jdamerow, line 45: This seems like business logic (checking that the app id is valid) and therefore does not belong in a controller. A hint is that you have the same code in the edit controller ;) 

Also, you haven't removed the problem of what happens if citesphere doesn't respond or respond slowly.  Now Quadriga would hang when the user clicks on save. I'm not sure what the best approach is. Maybe cache the results and use the cached data for this check here? It wouldn't guarantee that the apps are still valid but we could deal with the problem of deleted apps somewhere else.

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddCollectionController.java
```diff
@@ -19,19 +25,34 @@
...
-        collectionManager.addCollection(collectionForm.getName(), collectionForm.getDescription());
+        
+        List<CitesphereAppInfo> citesphereApps = citesphereConnector.getCitesphereApps();
+        HashSet<String> appSet = new HashSet<String>(
+                citesphereApps.stream().map(app -> app.getClientId()).collect(Collectors.toList()));
+        
+        for (String clientId : collectionForm.getApps()) {
+            if (!appSet.contains(clientId)) {
+                model.addAttribute("collectionForm", collectionForm);
+                return "auth/addCollection";
```
jdamerow: you probably want to show an error message in this case?

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/web/AddCollectionController.java
```diff
@@ -30,8 +31,14 @@ public String add(@Valid CollectionForm collectionForm, BindingResult result, Re
...
             return "auth/addCollection";
         }
-
-        collectionManager.addCollection(collectionForm.getName(), collectionForm.getDescription());
+        
+        try {
+            collectionManager.addCollection(collectionForm.getName(), collectionForm.getDescription(),
+                    collectionForm.getApps());
+        } catch (CitesphereAppNotFoundException e) {
+            result.rejectValue("apps", "error.collectionForm", e.getMessage());
```
jdamerow, line 42: stacktrace needs to be logged

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -81,14 +83,17 @@ public String edit(@PathVariable String id, @Valid CollectionForm collectionForm
...
+            collectionManager.editCollection(id, collectionForm.getName(), collectionForm.getDescription(), collectionForm.getApps());
             
             redirectAttributes.addFlashAttribute("alert_type", "success");
             redirectAttributes.addFlashAttribute("alert_msg", "Collection has been edited.");
             redirectAttributes.addFlashAttribute("show_alert", true);
             return "redirect:/auth/collections";
         } catch (InvalidObjectIdException | CollectionNotFoundException e) {
             return "error404Page";
+        } catch (CitesphereAppNotFoundException e) {
+            bindingResult.rejectValue("apps", "error.collectionForm", e.getMessage());
```
jdamerow, line 97: stacktrace needs to be logged

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -52,36 +61,36 @@ public Collection findCollection(String id) throws InvalidObjectIdException {
...
+            HashSet<String> appSet = new HashSet<String>(
+                    citesphereApps.stream().map(app -> app.getClientId()).collect(Collectors.toList()));
+            
+            if (apps != null) {
+                for (String clientId : apps) {
+                    if (!appSet.contains(clientId)) {
+                        throw new CitesphereAppNotFoundException("No Citesphere App found with client id " + clientId);
+                    }
+                }
+            }
```
jdamerow: this looks like duplciate code (same as in addcollection) and should be more DRY

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
There is something wrong with the checkboxes. I either can't select apps when creating a new collection (the checkboxes won't tick), or I was able to select on checkbox but now I can't unselect it.

jdamerow:
Make it so, Jenkins.

jdamerow:
still the same issue with checkboxes when adding a collection

jdamerow:
Make it so, Jenkins.

PR #21, opened 2021-10-08
Story/q20 18

Pooja444:

From the list of collections, it should be possible to click on a collection to see the collection page. That page should have the following info:

creation date

number of networks in collection

info about last submitted network to that collection (when, from whom)

a list of mappings for that collection with number triples in each mapping

default mapping

mappings that were submitted through endpoint or frontend
  

quadriga/pom.xml
```diff
@@ -270,6 +283,12 @@
             <version>3.0.4.RELEASE</version>
         </dependency>
 
+        <dependency>
+		    <groupId>org.thymeleaf.extras</groupId>
+		    <artifactId>thymeleaf-extras-java8time</artifactId>
+		    <version>3.0.4.RELEASE</version>
+		</dependency>
+
```
jdamerow, line 309: indentations in this file are off

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -11,6 +11,13 @@
     
     private String name;
     private String description;
+
+    // Time will be stored in UTC
```
jdamerow: this should be proper javadoc

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -11,6 +11,13 @@
...
+    private LocalDateTime creationTime;
```
jdamerow: this should use `OffsetDateTime`

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -1,19 +1,29 @@
...
 
+import edu.asu.diging.quadriga.api.v1.model.Context;
 import edu.asu.diging.quadriga.core.model.events.CreationEvent;
 
 public class EventGraph {
 
     private ObjectId _id;
     private CreationEvent  rootEvent;
     private DefaultMapping defaultMapping;
+    private LocalDateTime creationTime;
```
jdamerow: should use `OffsetDateTime`

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -1,19 +1,29 @@
...
+    private ObjectId collectionId;
+    private String appName;
```
jdamerow, line 17: what's this for?

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -1,19 +1,29 @@
...
+    private Context context;
     
-    public EventGraph() {}
+    public EventGraph() {
+        this.creationTime = LocalDateTime.now();
```
jdamerow: the creation time would come from the database, so it shouldn't be set here.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,95 @@
...
+            if(collection == null) {
+                return "error404Page";
+            }
+        } catch (InvalidObjectIdException e) {
+            return "error404Page";
+        }
+        
+        ObjectId collectionId = collection.getId();
+        
+        List<EventGraph> eventGraphs = eventGraphService.findAllEventGraphsByCollectionId(collectionId);
```
jdamerow, line 61: since here only the latest event graph is required, I would add a new manager method that gets only. the latest

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,95 @@
...
+     * This method gets all mappedTripleGroups that have the current collectionId for which
+     * the mappedTripleType is "DefaultMapping"
+     * 
+     * @param collectionId used to find mappedTripleGroupId
+     * @return the number of default mappings
+     */
+    private int getDefaultMappings(String collectionId) {
+        try {
+            List<MappedTripleGroup> defaultMappings = mappedTripleGroupService.findDefaultMappedTripleGroupsByCollectionId(collectionId.toString());
+            if(defaultMappings == null || defaultMappings.isEmpty()) return 0;
```
jdamerow: please follow our code style

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    /**
+     * This method gets all mappedTripleGroups that have the current collectionId for which
+     * the mappedTripleType is "DefaultMapping"
+     * 
+     * @param collectionId used to find mappedTripleGroupId
+     * @return the number of default mappings
+     */
+    private int getDefaultMappings(String collectionId) {
```
jdamerow: I'm confused, what does this method do? It's called getDefaultMappings but it returns an int?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            List<MappedTripleGroup> defaultMappings = mappedTripleGroupService.findDefaultMappedTripleGroupsByCollectionId(collectionId.toString());
+            if(defaultMappings == null || defaultMappings.isEmpty()) return 0;
+            return defaultMappings.size();
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
+            logger.error(ExceptionUtils.getStackTrace(e));
```
jdamerow: pass `e` into the logger

quadriga/src/main/java/edu/asu/diging/quadriga/web/EditCollectionController.java
```diff
@@ -40,17 +41,20 @@
...
+                CollectionForm collectionForm = new CollectionForm();
+                collectionForm.setId(id);
+                collectionForm.setName(collection.getName());
+                collectionForm.setDescription(collection.getDescription());
+                model.addAttribute("collectionForm", collectionForm);
+                return "auth/editCollection";
+            } else {
+                return "error404Page";
+            }
+        } catch (InvalidObjectIdException e) {
```
jdamerow, line 56: try/catch block should be only around the relevant code

jdamerow (changes requested):
there seem to be a lot of code unrelated to this ticket, so I only reviewed the parts that seemed to be related to this ticket.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
Pooja444: This is as per the ticket description - 
"info about last submitted network to that collection (when, from whom)"
For the "From whom" part, we need to display appName

**On slack we had this discussion:**

Julia Lessios-Damerow  4:08 PM
We probably need to add a story to get info about just one app. But the story Maulik works on would give you a list of all apps that you could also use. But I create a ticket for getting info about one app.
4:08
And sounds good, have a good evening.:+1:
Pooja Kulkarni  4:09 PM
Oh okay okay
So I’ll add a comment that once that ticket is done, app name should be displayed on collections page
Thanks and have a nice day!

Julia Lessios-Damerow  4:10 PM
:+1:


The "comment" I am referring to in the above discussion is the one I have added in the AddNetworkApiController:
`
            /**
             * This is a temporary solution as per the comment on story Q20-18
             * A new story will later be created to get info about just one app from citesphere using OAuth token.
             * This app's name should be stored in eventGraph instead of the client id
             */
`

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
Pooja444: Umm I am not sure I understand
So when a network is submitted, we add a new EventGraph entry in the database
The database doesn't have the EventGraph, we're adding a new one, not finding it from the DB

Did you mean that the creation time would come from the JSON request for the network submission?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
Pooja444: But the list of EventGraphs is also being used later in the Controller to get the number of submitted networks where one network can have multiple EventGraphs
Should I add two methods in the manager, one that gives latest event Graph and one that gives number of submitted networks?
I think this list is also being used for the next story (Q20-19)

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
Pooja444: As the logger won't accept 'e' as it the type of 'e' is Exception, I have to pass in e.getMessage(), is that fine?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
Pooja444: It returns number of default mappings
I'll update the method name to getNumberOfDefaultMappings

**This is as per our discussion on slack:**

Julia Lessios-Damerow  11:56 AM
The table would list all the mappings (right now only one, Default Mapping) and a number e.g. 10. Once we add functionality to add more mappings, the list would have eg Default Mappings (10), WiC Mappings (7), Test Mappings (12) etc

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
jdamerow: but this would refer to the app, right? so we need the id of the app or something? also, the variable name should indicate what is referred to here. Maybe "submittingApp" or something.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
jdamerow: I think the assumption is wrong that there are multiple eventGraphs for a network. There should be only one networks=eventGraph. and if I'm right, then yes, there should be one method to get the latest and one to count.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
jdamerow: I think this is an issue of confusing namings. A method named `getDefaultMappings` is expected to return default mappings. This method returns the number of triples in the default mapping group though, right?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
jdamerow: please check the different options. you can pass in a string and an exception.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
Pooja444: Actually as per the existing code, one network can have multiple EventGraphs -https://github.com/diging/quadriga-2.0/blob/develop/quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java#L54

What happens here is, for every "start" node in the network, an eventGraph is created
I also tested this logic with custom input and 3 EventGraphs were created by this logic for 1 network
I also verified this logic in old quadriga code, where, for displaying a network, multiple EventGraphs are merged by finding common nodes between the EventGraphs: https://github.com/diging/quadriga/blob/develop/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/EventParser.java

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
Pooja444: Yes, I have updated method name to getNumberOfDefaultMappings

Pooja444:
As this PR is dependent on Q20-12, will re-open it once Q20-12 is merged

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
jdamerow: Alright that makes sense, we don't care about the number of networks though. Just about how many event graphs there are for a collection. Networks are just to transfer event graphs to quadriga.

PR #24, opened 2022-01-31
Story/q20 18

Pooja444:

...Put description here...
  
From the list of collections, it should be possible to click on a collection to see the collection page. That page should have the following info:

creation date

number of networks in collection

info about last submitted network to that collection (when, from whom)

a list of mappings for that collection with number triples in each mapping

default mapping

mappings that were submitted through endpoint or frontend


quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -59,31 +59,43 @@
     @ResponseBody
     @RequestMapping(value = "/api/v1/collection/{collectionId}/network/add", method = RequestMethod.POST, consumes = MediaType.APPLICATION_JSON_VALUE)
     public HttpStatus processJson(@RequestBody Quadruple quadruple, @PathVariable String collectionId) {
+
+        // First we check whether a quadruple is present in request body
+        if (quadruple == null) {
+            logger.error("Quadruple not present in network submission request for collectionId: "  + collectionId);
+            return HttpStatus.NO_CONTENT;
```
jdamerow: this should return a bad request. No content means:
> The server has fulfilled the request but does not need to return an entity-body, and might want to return updated metainformation. The response MAY include new or updated metainformation in the form of entity-headers, which if present SHOULD be associated with the requested variant. [...]

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -59,31 +59,43 @@
...
         MappedTripleGroup mappedTripleGroup;
         try {
             mappedTripleGroup = mappedTripleGroupService.get(collectionId, MappedTripleType.DEFAULT_MAPPING);
             if(mappedTripleGroup == null) {
-                logger.error("Couldn't find or persist a new MappedTripleGroup entry for collectionId: " + collectionId);
                 return HttpStatus.NOT_FOUND;
             }
         } catch(InvalidObjectIdException | CollectionNotFoundException e)  {
-            logger.error("Couldn't submit network", e);
+            logger.error(e.getMessage());
```
jdamerow: that should not be changed

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -59,31 +59,43 @@
...
+            e.setContext(quadruple.getGraph().getMetadata().getContext());
+
+            /**
+             * A new story will later be created to get info about just one app from citesphere using OAuth token.
+             * This app's name should be stored in eventGraph instead of the client id
+             * Until that story is done, we need to store clientId instead of appName
+             * We can't store clientId yet as it depends on story Q20-3
+             * After merging story Q20-3, this needs to be changed to tokenInfo.getClientId()
+             */
+            e.setSubmittingApp("AppName");
```
jdamerow: There should be a `FIXME` in the comment

quadriga/src/main/java/edu/asu/diging/quadriga/config/MongoConfig.java
```diff
@@ -68,25 +73,34 @@ public MongoDatabaseFactory mongoDbFactory() throws UnknownHostException {
...
     @ReadingConverter
-    @Component
-    class OffsetDateTimeReadConverter implements Converter<String, OffsetDateTime> {
-        @Override
+    enum OffsetDateTimeReadConverter implements Converter<String, OffsetDateTime> {
+        INSTANCE;
         public OffsetDateTime convert(String source) {
             return OffsetDateTime.parse(source);
         }
     }
```
jdamerow, line 99: why are the above two classes turned into enums?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+    @RequestMapping(value = "/auth/collections/{id}", method = RequestMethod.GET)
+    public String get(@PathVariable String id, Model model) {
+        
+        Collection collection;
+        try {
+            collection = collectionManager.findCollection(id);
+            if(collection == null) {
+                return "error404Page";
+            }
+        } catch (InvalidObjectIdException e) {
```
jdamerow, line 54: needs to be logged

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+            return "error404Page";
+        }
+        
+        ObjectId collectionId = collection.getId();
+        
+        List<EventGraph> eventGraphs = eventGraphService.findAllEventGraphsByCollectionId(collectionId);
```
jdamerow: merge with lilne 61

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+            model.addAttribute("lastNetworkSubmittedBy", lastNetwork.getSubmittingApp());
+        }
+        
+        model.addAttribute("collectionName", collection.getName());
+        model.addAttribute("description", collection.getDescription());
+        model.addAttribute("creationTime", collection.getCreationTime().atZoneSameInstant(ZoneId.systemDefault()));
+        
+        // One network may have multiple eventGraphs, but all of them will have same sourceURI in the context
+        // This sourceURI will be used to group eventGraphs together that belong to the same network
+        model.addAttribute("numberOfSubmittedNetworks", eventGraphs.stream().collect(Collectors.groupingBy(event -> event.getContext().getSourceUri())).size());
```
jdamerow: this doesn't seem like a good approach. What if there are 10000 or more graphs in the db. This be done via MongoDb aggregations.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+     * This is because every default mapping has one predicate
+     * So, if the MappedTripleGroupId is present on n predicates, this collection
+     * must have n defaultMappings 
+     * 
+     * @param collectionId used to find mappedTripleGroupId
+     * @return the number of default mappings
+     */
+    private int getNumberOfDefaultMappings(String collectionId) {
+        try {
+            MappedTripleGroup mappedTripleGroup = mappedTripleGroupService.findByCollectionIdAndMappingType(collectionId, MappedTripleType.DEFAULT_MAPPING);
```
jdamerow: same as above, should be done via db

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+            if(mappedTripleGroup != null) {
+                List<Predicate> predicates = predicateManager.findByMappedTripleGroupId(mappedTripleGroup.get_id().toString());
+                if(predicates != null && !predicates.isEmpty()) {
+                    return predicates.size();
+                }
+            }
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
+            logger.error(e.getMessage());
```
jdamerow: needs to be properly logged

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
pkharge: @jdamerow , 
 This method is fetching mappedTripleGroups directly from mongo db, followed by fetching corresponding predicates from neo4j to get the total number of default mappings. Just want to confirm with you. Do I look at aggregating functions in neo4j to return the total count?

jdamerow: this probably refers to the way the number of predicates is retrieved. Getting the triple group then counting the predicates. Instead there should be a call to the db to get the number of predicates (so it doesns't load alll triples into memory).

jdamerow:
Codefactor failed.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
pkharge: Adding a FIXME raises an issue in codefactor check.

quadriga/src/main/java/edu/asu/diging/quadriga/config/MongoConfig.java
pkharge: Most of the default mongoConverters are enums. org.springframework.data.mongodb.core.convert.MongoConverters

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
jdamerow: That's fine in this case.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -1,17 +1,28 @@
...
+ *
+ */
 public class Collection {
 
     @Id
     private ObjectId _id;
     
     private String name;
     private String description;
+    private OffsetDateTime creationTime;
```
jdamerow, line 21: what about a setter for this variable?

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -1,17 +1,28 @@
...
     
+    public Collection() {
+        this.creationTime = OffsetDateTime.now();
```
jdamerow: Does this override any stored value or is the time set directly on the field when retrieved from the db?

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -1,20 +1,29 @@
...
 
     private ObjectId _id;
     private CreationEvent  rootEvent;
     private DefaultMapping defaultMapping;
+    private OffsetDateTime creationTime;
     private ObjectId collectionId;
-
-    public EventGraph() {}
+    private String submittingApp;
+    private Context context;
```
jdamerow: This should not be used here. The context is from the api package, model classes from the api should not be mixed with the stored model. If there is a context class needed it should be one from the model package.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -41,13 +50,32 @@ public void setDefaultMapping(DefaultMapping defaultMapping) {
         this.defaultMapping = defaultMapping;
     }
     
+    public OffsetDateTime getCreationTime() {
+        return creationTime;
+    }
```
jdamerow, line 50: What about the setter?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImpl.java
```diff
@@ -2,23 +2,63 @@
...
+import org.springframework.data.mongodb.core.query.Criteria;
 @Service
 public class EventGraphServiceImpl implements EventGraphService {
 
     @Autowired
     private EventGraphRepository repo;
-
+    
+    @Autowired
+    MongoTemplate mongoTemplate;
```
jdamerow: should be private

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImpl.java
```diff
@@ -2,23 +2,63 @@
...
+        Aggregation agg = newAggregation(unwind, match, group, count);
+        AggregationResults<Document> result = mongoTemplate.aggregate(
+                agg, mongoTemplate.getCollectionName(EventGraph.class), Document.class);
+
+        if (result.getMappedResults().isEmpty()) {
+            return 0;
+        }
+        return result.getMappedResults().get(0).getInteger("total");
+
+    }
```
jdamerow, line 73: This should be in a dedicated Mongo connection class. Otherwise, should we switch backends this class has to be changed rather than just replacing the dao class.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImpl.java
```diff
@@ -2,23 +2,63 @@
...
+        return repo.findByCollectionIdOrderByCreationTimeDesc(collectionId).orElse(null);
+    }
+
+    @Override
+    public long groupEventGraphsBySourceUri(ObjectId collectionId) {
+        UnwindOperation unwind = unwind("context");
+        GroupOperation group = Aggregation.group("context.sourceUri");
+        MatchOperation match= match(Criteria.where("collectionId").is(collectionId));
+        CountOperation count  = Aggregation.count().as("total");
+        Aggregation agg = newAggregation(unwind, match, group, count);
```
jdamerow: If I'm not wrong, this aggregation first groups all graphs by sourceuri and then filters by collection id, and then counts all groups not filtered out, right? but if two graphs were created for the same document, the source uri would be the same. This would give the wrong result, wouldn't it?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,118 @@
...
+    private EventGraphService eventGraphService;
+    
+    @Autowired
+    private MappedTripleGroupService mappedTripleGroupService;
+    
+    @Autowired
+    private PredicateManager predicateManager;
+    
+    @Autowired
+    MongoTemplate mongoTemplate;
```
jdamerow: a) should be private but b) controllers should almost never access data access layer classes. This shouldl be done from the service layer.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,118 @@
...
+            model.addAttribute("lastNetworkSubmittedAt", lastNetwork.getCreationTime().atZoneSameInstant(ZoneId.systemDefault()));
+            model.addAttribute("lastNetworkSubmittedBy", lastNetwork.getSubmittingApp());
+        }
+        
+        model.addAttribute("collectionName", collection.getName());
+        model.addAttribute("description", collection.getDescription());
+        model.addAttribute("creationTime", collection.getCreationTime().atZoneSameInstant(ZoneId.systemDefault()));
+        
+        // One network may have multiple eventGraphs, but all of them will have same sourceURI in the context
+        // This sourceURI will be used to group eventGraphs together that belong to the same network
```
jdamerow: does this comment belong here?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return the number of default mappings
+     */
+    private int getNumberOfDefaultMappings(String collectionId) {
+        try {
+            MappedTripleGroup mappedTripleGroup = mappedTripleGroupService.findByCollectionIdAndMappingType(collectionId, MappedTripleType.DEFAULT_MAPPING);
+            if(mappedTripleGroup != null) {    
+                return predicateManager.countPredicatesByMappedTripleGroup(mappedTripleGroup.get_id().toString());
+            }
+
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
```
jdamerow: exception needs to be logged

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+public class EventGraphServiceImplTest {
+
+    @Mock
+    private EventGraphRepository eventGraphRepository;
+
+    @InjectMocks
+    private EventGraphServiceImpl eventGraphServiceImpl;
+    
+    @Mock
+    MongoTemplate mongoTemplate;
```
jdamerow: should be private

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImpl.java
pkharge: This aggregation pipeline first filters based on collection id, then groups based on sourceUri. count returns the total count of remaining documents in the pipeline. This basically gives the number of distinct sourceUris for a given collection id.

jdamerow: yes, but source uris are specific to a text. so three event graphs might have the same source uri if they were all created for the same text. This method however, shoulld return the count of all event graphs in a collelction, right?

also, I"m just noticing, the method should be called `countEventGraphs` or something.

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/EventGraphRepository.java
```diff
@@ -1,10 +1,15 @@
...
+import java.util.Optional;
+
 import org.bson.types.ObjectId;
 import org.springframework.data.mongodb.repository.MongoRepository;
 
 import edu.asu.diging.quadriga.core.model.EventGraph;
 
 public interface EventGraphRepository extends MongoRepository<EventGraph, ObjectId> {
 
+    public Optional<List<EventGraph>> findFirstByCollectionIdOrderByCreationTimeDesc(ObjectId collectionId);
```
pkharge: Note: Have modified this to findFirst since we just need the latest event graph.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
pkharge: This constructor is only called while adding a new collection.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImpl.java
```diff
@@ -2,23 +2,58 @@
...
 
+
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.aggregation.Aggregation;
+import org.springframework.data.mongodb.core.aggregation.AggregationResults;
+import org.springframework.data.mongodb.core.aggregation.CountOperation;
+import org.springframework.data.mongodb.core.aggregation.GroupOperation;
+import org.springframework.data.mongodb.core.aggregation.MatchOperation;
+import org.springframework.data.mongodb.core.aggregation.UnwindOperation;
+import org.springframework.data.mongodb.core.query.Criteria;
```
jdamerow: remove unnecessary imports

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/EventGraphRepository.java
jdamerow: I believe with `findFirst` you can just return `Optional<EventGraph>`?

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (InvalidObjectIdException e) {
+            logger.error("Couldn't find collection ", e);
+            return "error404Page";
+        }
+        
+       
+        
+        List<EventGraph> eventGraph = eventGraphService.findLatestEventGraphByCollectionId(collection.getId());
+        
+        if(!eventGraph.isEmpty()) {
```
jdamerow: since `findLatestEventGraphByCollectionId` might return null, this would throw an exception in the case of null.

jdamerow (changes requested):

jdamerow:
Also, I'm not a huge fan of setting creation time in the constructor. The creation time does not necessarily mean the time an object was created but the time the thing the object represents was created. There could even be several objects representing the same thing. I don't see that it would create any issues right now but I can think of cases where it might create issues down the road. Setting creation time is really more business logic than object logic, so i would set creation times in the class creating the objects.

jdamerow:
Make it so, Jenkins.

jdamerow:
to that point actually, since existing networks/collections don't have a creation time, it now shows creation time is the time the object was retrieved from the database. It should just be blank. It's not an issue since we don't have a production version yet, but I wouldn't be surprised if we run into similar issues at some point.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/PredicateManagerImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    private PredicateRepository predicateRepository;
+
+    @Override
+    public List<Predicate> findByMappedTripleGroupId(String mappedTripleGroupId) {
+        return predicateRepository.findByMappedTripleGroupId(mappedTripleGroupId).orElse(null);
+
+    }
+
+    @Override
+	public int countPredicatesByMappedTripleGroup(String mappedTripleGroupId) {
```
jdamerow: indentation

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
on collection page:
```
java.lang.NullPointerException
	edu.asu.diging.quadriga.web.DisplayCollectionController.get(DisplayCollectionController.java:72)
	java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	java.base/java.lang.reflect.Method.invoke(Method.java:566)
	org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:190)
	org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:138)
	org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:105)
	org
```

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,112 @@
...
+            logger.error("Couldn't find collection ", e);
+            return "error404Page";
+        }
+        
+       
+        
+        EventGraph latestNetwork = eventGraphService.findLatestEventGraphByCollectionId(collection.getId());
+        
+        if( latestNetwork!=null ) {
+            model.addAttribute("lastNetworkSubmittedAt", latestNetwork.getCreationTime().atZoneSameInstant(ZoneId.systemDefault()));
```
jdamerow: actually, format the date on the front end. the user wants to see their time not the server time (e.g.  I would currently see a timestamp in the future).

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/EventGraph.java
```diff
@@ -9,12 +11,17 @@ public class EventGraph {
...
-    public EventGraph() {}
+    private String submittingApp;
     
+    public EventGraph() {
+        this.creationTime = OffsetDateTime.now();
+    }
+
     public EventGraph(CreationEvent root) {
         this.rootEvent = root;
+        this.creationTime = OffsetDateTime.now();
```
jdamerow: Let's take this out of here same, as for the collection classes.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,114 @@
...
+        
+        if( latestNetwork!=null ) {
+            model.addAttribute("lastNetworkSubmittedAt", latestNetwork.getCreationTime());
+            model.addAttribute("lastNetworkSubmittedBy", latestNetwork.getSubmittingApp());
+        }
+        
+        model.addAttribute("collectionName", collection.getName());
+        model.addAttribute("description", collection.getDescription());
+                
+        model.addAttribute("creationTime", collection.getCreationTime());
```
jdamerow: why don't you just put the collection into the model? then you have access to any properties you need in the template.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+        Mockito.when(mongoTemplate.getCollectionName(eq(EventGraph.class))).thenReturn("EVENT_GRAPH");
+        Mockito.when(mongoTemplate.aggregate(any(Aggregation.class), any(String.class), eq(Document.class)))
+            .thenReturn(aggregationResultsMock);
+        
+        long totalCount = eventGraphDaoImpl.countEventGraphsByCollectionId(collectionObjectId);
+
+        //Both the event graphs belong to the group since the sourceUri is same. 
+        Assert.assertEquals(totalCount, 2);
+    }
+
```
jdamerow, line 149: another test cases for 0 event graphs is missing.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -59,31 +60,48 @@ public class AddNetworkApiController {
...
-        if (quadruple == null) {
-            return HttpStatus.NO_CONTENT;
-        }
-
         // save network
         List<CreationEvent> events = networkMapper.mapNetworkToEvents(quadruple.getGraph());
-        List<EventGraph> eventGraphs = events.stream().map(e -> new EventGraph(e)).collect(Collectors.toList());
+        List<EventGraph> eventGraphs = events.stream().map(e ->  {
+            EventGraph eventGraph = new EventGraph(e);
+            eventGraph.setCreationTime(OffsetDateTime.now());
```
jdamerow: this should be done in the manager class that saves the graph

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiController.java
```diff
@@ -59,31 +60,48 @@ public class AddNetworkApiController {
...
+             * FIXME:
+             * 
+             * A new story will later be created to get info about just one app from citesphere using OAuth token.
+             * This app's name should be stored in eventGraph instead of the client id
+             * Until that story is done, we need to store clientId instead of appName
+             * We can't store clientId yet as it depends on story Q20-3
+             * After merging story Q20-3, this needs to be changed to tokenInfo.getClientId()
+             */
+            e.setSubmittingApp("AppName");
         });
```
jdamerow: Actually, the whole logic should be moved into a manager method. I would create a method that takes a graph object, maps it to CreationEvent, and then creates event graphs from it.

jdamerow: It's too much business logic here for a controller.

quadriga/src/main/java/edu/asu/diging/quadriga/legacy/service/impl/RepositoryManager.java
```diff
@@ -38,7 +39,12 @@ public List<String> processXMLandStoretoDb(String xml, String type) throws URISy
 
         creationEventList = xmlToObject.parseXML(xml);
         List<EventGraph> flattenedlist = creationEventList.stream().flatMap(List::stream)
-                .map(event -> new EventGraph(event)).collect(Collectors.toList());
+                .map(event -> {
+                    EventGraph eventGraph = new EventGraph(event);
+                    eventGraph.setCreationTime(OffsetDateTime.now());
```
jdamerow: once this is moved into the manager, this line isn't needed here.

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,109 @@
...
+            return "error404Page";
+        }
+        
+       
+        
+        EventGraph latestNetwork = eventGraphService.findLatestEventGraphByCollectionId(collection.getId());
+        
+        if( latestNetwork!=null ) {
+            model.addAttribute("lastNetworkSubmittedAt", latestNetwork.getCreationTime());
+            model.addAttribute("lastNetworkSubmittedBy", latestNetwork.getSubmittingApp());
```
jdamerow: same as before, justs put the latest network into them model.

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/EventGraphServiceImplTest.java
```diff
@@ -0,0 +1,150 @@
...
+        
+        long totalCount = eventGraphDaoImpl.countEventGraphsByCollectionId(collectionObjectId);
+
+        //Both the event graphs belong to the group. 
+        Assert.assertEquals(totalCount, 2);
+    }
+    
+    
+    @Test
+    public void test_countZeroEventGraphsBy_success() throws InterruptedException {
```
jdamerow: please read our guidelines for how to name test methods.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/web/DisplayCollectionController.java
```diff
@@ -0,0 +1,80 @@
...
+            logger.error("Couldn't find collection ", e);
+            return "error404Page";
+        }
+        
+       
+        
+        EventGraph latestNetwork = eventGraphService.findLatestEventGraphByCollectionId(collection.getId());
+        
+        if( latestNetwork!=null ) {
+            model.addAttribute("latestNetwork",latestNetwork);
```
jdamerow: if latestNetwork is null the result is pretty much the same, isn't it? Thymeleaf won't find anything when it looks for `latestNetwork` in the model. So, do you really need this if since you need to check for this property in the model anyways before accessing it?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
after this should be good!

jdamerow:
Make it so, Jenkins.

PR #26, opened 2022-05-17
Story/q20 13

MaulikLim:

[Q20-13] - there needs to be an api endpoint to get all collection
  
... describe here ...


quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionsApiController.java
```diff
@@ -0,0 +1,33 @@
...
+import edu.asu.diging.quadriga.core.service.CollectionManager;
+
+@Controller
+public class GetCollectionsApiController {
+
+    @Autowired
+    private CollectionManager collectionManager;
+
+    @GetMapping("/api/v1/collections")
+    ResponseEntity<List<Collection>> getCollections(
```
jdamerow: should be public

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionsApiController.java
```diff
@@ -0,0 +1,33 @@
...
+            @RequestHeader(name = "Authorization", required = false) String authHeader, Authentication authentication) {
+        if (!(authentication.getDetails() instanceof TokenInfo)) {
+            return new ResponseEntity<>(HttpStatus.OK);
+        }
+        TokenInfo tokenInfo = (TokenInfo) authentication.getDetails();
```
jdamerow: I would try to put this logic into an aspect since this will be required by more than just this one controller in the future. The aspect should inject the tokeninfo on successful authorization and otherwise return unauthorized for now.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionsApiController.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+
+import edu.asu.diging.quadriga.aspect.annotation.InjectToken;
+import edu.asu.diging.quadriga.config.web.TokenInfo;
+import edu.asu.diging.quadriga.core.model.Collection;
+import edu.asu.diging.quadriga.core.service.CollectionManager;
+
+@Controller
+public class GetCollectionsApiController {
```
jdamerow: Let's call this `CollectionsApiControlller` and if we need other CRUD methods we can put them in here.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,52 @@
...
+        } else if (tokenIndex >= 0) {
+            args[tokenIndex] = null;
+        }
+        return joinPoint.proceed(args);
+    }
+
+    private int getIndexOf(Object[] args, Class<?> classType) {
+        for (int i = 0; i < args.length; i++) {
+            if (classType.isAssignableFrom(args[i].getClass())) {
+                return i;
```
jdamerow: why don't you return `args[i]` here?

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,52 @@
...
+    @Pointcut("execution(* *(.., edu.asu.diging.quadriga.config.web.TokenInfo, ..))")
+    public void tokenParamMethod() {}
+
+    @Around("annotatedMethod() && authenticationParamMethod() && tokenParamMethod()")
+    public Object authAdvice(ProceedingJoinPoint joinPoint) throws Throwable {
+        Object[] args = joinPoint.getArgs();
+        int authIndex = getIndexOf(args, Authentication.class);
+        int tokenIndex = getIndexOf(args, TokenInfo.class);
+
+        if (authIndex >= 0 && tokenIndex >= 0) {
```
jdamerow: if you don't use indexes (see comment below) you simply check for null here.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
jdamerow: it will make above code more concise

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
MaulikLim: I need the tokenIndex to replace the existing object in args array. So I might have to find the index either way.

MaulikLim:
I have refactored the pointcut expression to only check for the TokenInfo parameter as it didn't behave appropriately for the scenario when the Authentication parameter is missing and TokenInfo is present. In that case, the aspect is not wired to the method and so the controller receives an empty TokenInfo object instead of null. This makes the API return collections with no apps configured instead of Unauthorized access response.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,48 @@
...
+    public Object authAdvice(ProceedingJoinPoint joinPoint) throws Throwable {
+        Object[] args = joinPoint.getArgs();
+        int authIndex = getIndexOf(args, Authentication.class);
+        int tokenIndex = getIndexOf(args, TokenInfo.class);
+
+        if (tokenIndex >= 0) {
+            args[tokenIndex] = null;
+            if (authIndex >= 0) {
+                Authentication authentication = (Authentication) args[authIndex];
+                if (authentication != null && authentication.getDetails() instanceof TokenInfo) {
```
jdamerow: authentication can never be null, can it? if it were authIndex would be -1?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/CollectionsApiControlller.java
```diff
@@ -0,0 +1,32 @@
...
+@Controller
+public class CollectionsApiControlller {
+
+    @Autowired
+    private CollectionManager collectionManager;
+
+    @InjectToken
+    @GetMapping("/api/v1/collections")
+    public ResponseEntity<List<Collection>> getCollections(Authentication authentication, TokenInfo tokenInfo) {
+        if (tokenInfo == null) {
```
jdamerow: when does this happen? The aspect should take care of this or we would have to add that to every api controller.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,54 @@
...
+                if (authentication.getDetails() instanceof TokenInfo) {
+                    args[tokenIndex] = authentication.getDetails();
+                }
+            }
+
+            if (args[tokenIndex] == null) {
+                return new ResponseEntity<>(HttpStatus.UNAUTHORIZED);
+            }
+        }
+        return joinPoint.proceed(args);
```
jdamerow: This should never actually happen since the aspect is only run on methods that have a TokenInfo parameter, shouldn't it?

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,54 @@
...
+@Configuration
+public class AuthAspect {
+
+    @Pointcut("execution(@edu.asu.diging.quadriga.aspect.annotation.InjectToken * *(..))")
+    public void annotatedMethod() {}
+
+    @Pointcut("execution(* *(.., edu.asu.diging.quadriga.config.web.TokenInfo, ..))")
+    public void tokenParamMethod() {}
+
+    @Around("annotatedMethod() && tokenParamMethod()")
```
jdamerow: since those two go hand in hand (annotating the method only makes sense if there is a token parameter), I would raise a runtime exception if there is no token param.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,54 @@
...
+    public Object authAdvice(ProceedingJoinPoint joinPoint) throws Throwable {
+        Object[] args = joinPoint.getArgs();
+        int authIndex = getIndexOf(args, Authentication.class);
+        int tokenIndex = getIndexOf(args, TokenInfo.class);
+
+        if (tokenIndex >= 0) {
+            args[tokenIndex] = null;
+            if (authIndex >= 0) {
```
jdamerow: actually, I think in this case I would use the `SecurityContextHolder` to get the current authentication. Then you don't have to search through the arguments and it's one thing less the developer has to remember.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,54 @@
...
+                Authentication authentication = (Authentication) args[authIndex];
+                if (authentication.getDetails() instanceof TokenInfo) {
```
jdamerow: if not, throw runtime exception

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
MaulikLim: Ideally, yes. The annotation should only be used on APIs which are secured and the authentication has TokenInfo. I was thinking of case where, somehow the request is authenticated through some other means (maybe another security config) but does not actually have TokenInfo -> App Client Id.

jdamerow: but then it wouldn't be annotated with `@InjectToken` would it?

MaulikLim: Yes. I'm only covering for an incorrect annotation usage.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,46 @@
...
+    @Around("annotatedMethod() && tokenParamMethod()")
+    public Object authAdvice(ProceedingJoinPoint joinPoint) throws Throwable {
+        Object[] args = joinPoint.getArgs();
+        int tokenIndex = getIndexOf(args, TokenInfo.class);
+
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        if (authentication != null && authentication.getDetails() instanceof TokenInfo) {
+            args[tokenIndex] = authentication.getDetails();
+        } else {
+            throw new RuntimeException("No token info found to retrieve app client id");
```
jdamerow: Let's create our own exception here that we can reuse in other places.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,46 @@
...
+public class AuthAspect {
+
+    @Pointcut("execution(@edu.asu.diging.quadriga.aspect.annotation.InjectToken * *(..))")
+    public void annotatedMethod() {}
+
+    @Pointcut("execution(* *(.., edu.asu.diging.quadriga.config.web.TokenInfo, ..))")
+    public void tokenParamMethod() {}
+
+    @Around("annotatedMethod() && tokenParamMethod()")
+    public Object authAdvice(ProceedingJoinPoint joinPoint) throws Throwable {
```
jdamerow: needs javadoc

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

quadriga/src/main/java/edu/asu/diging/quadriga/core/exceptions/TokenNotFoundException.java
```diff
@@ -0,0 +1,23 @@
+package edu.asu.diging.quadriga.core.exceptions;
+
+public class TokenNotFoundException extends Exception {
```
jdamerow: That should be a runtime exception.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
jdamerow: does this method need to throw `Throwable`?

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,59 @@
...
+import edu.asu.diging.quadriga.config.web.TokenInfo;
+import edu.asu.diging.quadriga.core.exceptions.TokenNotFoundException;
+
+/**
+ * Aspect configuration class injects the token info from citesphere into the appropriate method parameter
+ * @author Maulik Limbadiya
+ *
+ */
+@Aspect
+@Configuration
```
jdamerow: aspects are not configurations. They should be annotated with component.

quadriga/src/main/java/edu/asu/diging/quadriga/aspect/AuthAspect.java
```diff
@@ -0,0 +1,59 @@
+package edu.asu.diging.quadriga.aspect;
```
jdamerow: `aspect` package should be part of the `core` package

jdamerow (changes requested):

MaulikLim:
`Throwable` is required for the aspect method. It does not work if I just add throws `TokenNotFoundException`. Gives compilation error.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/CollectionsApiControlller.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+
+import edu.asu.diging.quadriga.config.web.TokenInfo;
+import edu.asu.diging.quadriga.core.aspect.annotation.InjectToken;
+import edu.asu.diging.quadriga.core.model.Collection;
+import edu.asu.diging.quadriga.core.service.CollectionManager;
+
+@Controller
+public class CollectionsApiControlller {
```
jdamerow: There is a spelling mistake in the class name.

jdamerow (changes requested):
There should be also a pull request in Postman describing the new endpoint (I've already created the request, it just needs a description).

jdamerow:
Make it so, Jenkins.

jdamerow:
After that should be good.

jdamerow:
Make it so, Jenkins.

PR #28, opened 2022-06-29
Story/q20 14

MaulikLim:

[Q20-14] - There needs to be an endpoint to get all mapped triples for a collection
  
... describe here ...


quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionTriplesApiController.java
```diff
@@ -0,0 +1,78 @@
...
+    private CollectionManager collectionManager;
+
+    @Autowired
+    private MappedTripleService mappedTripleService;
+
+    @Autowired
+    private MappedTripleGroupService mappedTripleGroupService;
+
+    @InjectToken
+    @GetMapping(value = "/api/v1/collection/{collectionId}/triples", produces = MediaType.APPLICATION_JSON_VALUE)
```
jdamerow: let's make this `.../triples/{tripleGroupId}` and then only return the default mappings. then we can change it later (would actually not take much).

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionTriplesApiController.java
```diff
@@ -0,0 +1,78 @@
...
+    public ResponseEntity<String> getTriples(@PathVariable("collectionId") String collectionId, TokenInfo tokenInfo) {
+        try {
+
+            Collection collection = collectionManager.findCollection(collectionId);
+
+            if (collection.getApps() == null || collection.getApps().isEmpty()
+                    || !collection.getApps().contains(tokenInfo.getClient_id())) {
+                return new ResponseEntity<>(HttpStatus.UNAUTHORIZED);
+            }
```
jdamerow: that shouldn't stay the only method that access collection and needs a permission check, so it should be in an aspect.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionTriplesApiController.java
```diff
@@ -0,0 +1,78 @@
...
+                            mappedTripleService.getMappedTriples(mappedTripleGroup.get_id().toString())),
+                    HttpStatus.OK);
+
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
+            logger.error("No collection found for id {}", collectionId);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (JsonProcessingException e) {
+            logger.error("Error while converting the response to JSON");
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
```
jdamerow, line 88: the try/catch blocks should be around the code that's throwing it not all of it. And exceptions need to be logged.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Triple.java
```diff
@@ -0,0 +1,40 @@
+package edu.asu.diging.quadriga.core.model;
+
+/**
+ * Holds the data of a mapped triple extracted from the submitted network.
+ * @author Maulik Limbadiya
+ *
+ */
+public class Triple {
```
jdamerow: what's the difference to `DefaultMapping`?

jdamerow (changes requested):
This doesn't seem to be done:
> The results need to be paginated (the client will need to make multiple calls to get all data).

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionTriplesApiController.java
```diff
@@ -0,0 +1,88 @@
...
+            TokenInfo tokenInfo) {
+
+        int pageInt = 1;
+        int pageSizeInt = 10;
+
+        try {
+            pageInt = Integer.parseInt(page);
+            pageSizeInt = Integer.parseInt(pageSize);
+        } catch (NumberFormatException e) {
+            logger.warn("Trying to access invalid page number: " + page);
```
jdamerow: exception needs to be logged

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GetCollectionTriplesApiController.java
```diff
@@ -0,0 +1,88 @@
...
+
+        }
+
+        MappedTripleGroup mappedTripleGroup;
+        try {
+            mappedTripleGroup = mappedTripleGroupService.findByCollectionIdAndMappingType(collectionId,
+                    MappedTripleType.DEFAULT_MAPPING);
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
+            logger.error("No collection found for id {}", collectionId);
```
jdamerow: exception needs to be logged

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
besides that just the minor change to the documentation.

jdamerow:
Make it so, Jenkins.

PR #29, opened 2022-07-18
Story/q20 23

MaulikLim:

[Q20-23] - Deleting of collections
  
Need to update existing collection entries in database with the new field.
`db.collection.updateMany({}, { $set: { "isArchived" : false } });`

[Q20-23]: https://diging.atlassian.net/browse/Q20-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/quadriga/pom.xml
+++ b/quadriga/pom.xml
@@ -21,8 +21,8 @@
 
         <db.driver>com.mysql.jdbc.Driver</db.driver>
         <db.database.url>jdbc:mysql://localhost:3306/quadriga</db.database.url>
-        <db.user></db.user>
-        <db.password></db.password>
+        <db.user>db-user</db.user>
+        <db.password>db-password</db.password>
 
         <email.user></email.user>
         <email.password></email.password>
@@ -39,16 +39,14 @@
         <mongo.database.authdb>admin</mongo.database.authdb>
 
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
-        <neo4j.database></neo4j.database>
+        <neo4j.database>quadriga2</neo4j.database>
         
-        <citesphere.base.url></citesphere.base.url>
-        <citesphere.client.id></citesphere.client.id>
-        <citesphere.client.secret></citesphere.client.secret>
-        
-        <conceptpower.base.url></conceptpower.base.url>
-        <conceptpower.id.endpoint></conceptpower.id.endpoint>
-
-
+        <citesphere.base.url>https://diging-dev.asu.edu/citesphere-review/</citesphere.base.url>
+        <citesphere.client.id>OAUTHCLIENT40</citesphere.client.id>
+        <citesphere.client.secret>1fef2d29-7d30-435e-9210-c56eb18131cf</citesphere.client.secret>
+        <!--The base.url can have trailing '/' -->
+        <conceptpower.base.url>https://chps.asu.edu/conceptpower/rest/</conceptpower.base.url>
+        <conceptpower.id.endpoint>https://chps.asu.edu/conceptpower/</conceptpower.id.endpoint>
 
         <log.level>info</log.level>
 
--- a/quadriga/pom.xml
+++ b/quadriga/pom.xml
@@ -21,8 +21,8 @@
 
         <db.driver>com.mysql.jdbc.Driver</db.driver>
         <db.database.url>jdbc:mysql://localhost:3306/quadriga</db.database.url>
-        <db.user>db-user</db.user>
-        <db.password>db-password</db.password>
+        <db.user></db.user>
+        <db.password></db.password>
 
         <email.user></email.user>
         <email.password></email.password>
@@ -39,11 +39,11 @@
         <mongo.database.authdb>admin</mongo.database.authdb>
 
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
-        <neo4j.database>quadriga2</neo4j.database>
+        <neo4j.database>quadriga</neo4j.database>
         
-        <citesphere.base.url>https://diging-dev.asu.edu/citesphere-review/</citesphere.base.url>
-        <citesphere.client.id>OAUTHCLIENT40</citesphere.client.id>
-        <citesphere.client.secret>1fef2d29-7d30-435e-9210-c56eb18131cf</citesphere.client.secret>
+        <citesphere.base.url></citesphere.base.url>
+        <citesphere.client.id></citesphere.client.id>
+        <citesphere.client.secret></citesphere.client.secret>
         <!--The base.url can have trailing '/' -->
         <conceptpower.base.url>https://chps.asu.edu/conceptpower/rest/</conceptpower.base.url>
         <conceptpower.id.endpoint>https://chps.asu.edu/conceptpower/</conceptpower.id.endpoint>
--- a/quadriga/pom.xml
+++ b/quadriga/pom.xml
@@ -20,9 +20,9 @@
         <admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
 
         <db.driver>com.mysql.jdbc.Driver</db.driver>
-        <db.database.url>jdbc-url-to-db</db.database.url>
-        <db.user>db-user</db.user>
-        <db.password>db-password</db.password>
+        <db.database.url>jdbc:mysql://localhost:3306/quadriga</db.database.url>
+        <db.user></db.user>
+        <db.password></db.password>
 
         <email.user></email.user>
         <email.password></email.password>
@@ -39,11 +39,16 @@
         <mongo.database.authdb>admin</mongo.database.authdb>
 
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
-        <neo4j.database>quadriga</neo4j.database>
+        <neo4j.database></neo4j.database>
         
         <citesphere.base.url></citesphere.base.url>
         <citesphere.client.id></citesphere.client.id>
         <citesphere.client.secret></citesphere.client.secret>
+        
+        <conceptpower.base.url></conceptpower.base.url>
+        <conceptpower.id.endpoint></conceptpower.id.endpoint>
+
+
 
         <log.level>info</log.level>
 
--- a/quadriga/pom.xml
+++ b/quadriga/pom.xml
@@ -20,9 +20,9 @@
         <admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
 
         <db.driver>com.mysql.jdbc.Driver</db.driver>
-        <db.database.url>jdbc:mysql://localhost:3306/quadriga</db.database.url>
-        <db.user></db.user>
-        <db.password></db.password>
+        <db.database.url>jdbc-url-to-db</db.database.url>
+        <db.user>db-user</db.user>
+        <db.password>db-password</db.password>
 
         <email.user></email.user>
         <email.password></email.password>
@@ -44,9 +44,6 @@
         <citesphere.base.url></citesphere.base.url>
         <citesphere.client.id></citesphere.client.id>
         <citesphere.client.secret></citesphere.client.secret>
-        <!--The base.url can have trailing '/' -->
-        <conceptpower.base.url>https://chps.asu.edu/conceptpower/rest/</conceptpower.base.url>
-        <conceptpower.id.endpoint>https://chps.asu.edu/conceptpower/</conceptpower.id.endpoint>
 
         <log.level>info</log.level>
 
--- a/quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
+++ b/quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
@@ -19,6 +19,7 @@ public class Collection {
     private String name;
     private String description;
     private OffsetDateTime creationTime;
+    private boolean isArchived;
 
     /**
      * List of Client Ids for Citesphere Apps which are associated with the collection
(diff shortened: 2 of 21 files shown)


quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImpl.java
```diff
@@ -100,8 +105,13 @@ public Collection editCollection(String id, String name, String description, Lis
     public void deleteCollection(String id) throws CollectionNotFoundException, InvalidObjectIdException {
         Collection collection = findCollection(id);
         
-        if(Objects.nonNull(collection)) {
-            
+        if (collection != null) {
+            EventGraph eventGraph = eventGraphService.findLatestEventGraphByCollectionId(new ObjectId(id));
+            if (eventGraph != null) {
```
jdamerow, line 114: there needs to be a comment here why this check is done

quadriga/src/main/java/edu/asu/diging/quadriga/web/ListArchivedCollectionController.java
```diff
@@ -0,0 +1,33 @@
...
+
+import edu.asu.diging.quadriga.core.data.CollectionRepository;
+
+@Controller
+public class ListArchivedCollectionController {
+
+    @Autowired
+    private CollectionRepository collectionRepo;
+
+    @RequestMapping(value = "/auth/archived-collections", method = RequestMethod.GET)
```
jdamerow: should be `/auth/collections/archived`

quadriga/src/main/java/edu/asu/diging/quadriga/web/ListArchivedCollectionController.java
```diff
@@ -0,0 +1,33 @@
...
+    public String list(@RequestParam(value = "page", required = false, defaultValue = "1") String page,
+            @RequestParam(value = "size", required = false, defaultValue = "20") String size, Model model) {
+
+        int pageInt = 0;
+        int sizeInt = 20;
+
+        pageInt = Integer.parseInt(page) - 1;
+        sizeInt = Integer.parseInt(size);
```
jdamerow: there should be try/caches around the parsing so the variables are simply set to the default values if bogus values are passed.

jdamerow (changes requested):

quadriga/src/main/webapp/WEB-INF/views/auth/displayCollection.html
```diff
@@ -4,7 +4,11 @@
 </head>
 <body>
 	<div layout:fragment="content">
-		<a th:href="@{/auth/collections}" class="btn btn-primary btn-sm">
+		<a th:if="${collection.isArchived()}" th:href="@{/auth/archived-collections}" class="btn btn-primary btn-sm">
```
jdamerow: I believe this link is wrong now?

quadriga/src/main/webapp/WEB-INF/views/auth/showArchivedCollections.html
```diff
@@ -0,0 +1,49 @@
+<html layout:decorate="~{layouts/main}">
+<head>
+<title>Quadriga 2.0</title>
+</head>
+<body>
+    <div layout:fragment="content">
+        <a class ="btn btn-primary btn-nav" th:href="@{/auth/collections}"><i class="fas fa-arrow-left" aria-hidden="true"></i> Back</a>
+        <h2>Archived Collections</h2>
+        <div th:if="${collections.empty}">No Archived Collections available</div>
+        <div action="#" th:action="@{/auth/archived-collections}" method="GET">
```
jdamerow: same here

quadriga/src/main/webapp/WEB-INF/views/auth/showcollections.html
```diff
@@ -5,6 +5,9 @@
 <body>
 	<div layout:fragment="content">
 		<h2>My Collections</h2>
+        <a th:href="@{/auth/archived-collections}">
```
jdamerow: and here

jdamerow (changes requested):

quadriga/src/main/webapp/WEB-INF/views/auth/showcollections.html
MaulikLim: fixed

jdamerow:
Make it so, Jenkins.

jdamerow:
There should be a message when deleting a collection with networks that the collection will be archived not deleted.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/Collection.java
```diff
@@ -19,6 +19,7 @@ public class Collection {
     private String name;
     private String description;
     private OffsetDateTime creationTime;
+    private boolean isArchived;
```
jdamerow: variable name should be `archived`

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -132,6 +138,29 @@ public boolean matches(Collection arg0) {
...
+        collection.setId(id);
+        collection.setName(name);
+        collection.setDescription(desc);
+        collection.setArchived(true);
+        
+        Mockito.when(collectionRepo.findById(id)).thenReturn(Optional.of(collection));
+        Mockito.when(eventGraphService.findLatestEventGraphByCollectionId(id)).thenReturn(new EventGraph());
+        Mockito.when(collectionRepo.save(Mockito.any())).thenReturn(response);
+        
+        managerToTest.deleteCollection(id.toString());
```
jdamerow: there need to be checks after this method is run to make sure it does what it's supposed to be doing.

jdamerow (changes requested):

quadriga/pom.xml
```diff
@@ -20,9 +20,9 @@
         <admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
 
         <db.driver>com.mysql.jdbc.Driver</db.driver>
-        <db.database.url>jdbc-url-to-db</db.database.url>
-        <db.user>db-user</db.user>
-        <db.password>db-password</db.password>
+        <db.database.url>jdbc:mysql://localhost:3306/quadriga</db.database.url>
+        <db.user></db.user>
+        <db.password></db.password>
```
jdamerow: this should not be changed (database url is ok, but the rest should stay the way it was)

quadriga/pom.xml
```diff
@@ -39,11 +39,16 @@
...
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
-        <neo4j.database>quadriga</neo4j.database>
+        <neo4j.database></neo4j.database>
         
         <citesphere.base.url></citesphere.base.url>
         <citesphere.client.id></citesphere.client.id>
         <citesphere.client.secret></citesphere.client.secret>
+        
+        <conceptpower.base.url></conceptpower.base.url>
+        <conceptpower.id.endpoint></conceptpower.id.endpoint>
```
jdamerow: add a comment about leading/trailing slashes here (e.g. should it be `https://chps.asu.edu/conceptpower/rest/` or `https://chps.asu.edu/conceptpower/rest`?)

quadriga/src/main/resources/log4j.xml
```diff
@@ -48,7 +48,7 @@
     <root>
         <priority value="warn" />
         <appender-ref ref="console" />
-        <appender-ref ref="FILE" />
+        <!-- <appender-ref ref="FILE" />-->
```
jdamerow: should not be commented out

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -132,6 +140,41 @@ public boolean matches(Collection arg0) {
...
+        }
+        catch (CollectionNotFoundException e)
+        {
+            fail("Exception thrown during test execution: " + e.getMessage());
+            
+        }
+        catch(InvalidObjectIdException e)
+        {
+            fail("Exception thrown during test execution: " + e.getMessage());
+        }
```
jdamerow: I don't think you need this, just let it throw the exceptions and the tests will fail automatically.

jdamerow: when does the test pass though? There needs to be an assert statement.

jdamerow (changes requested):

quadriga/logs/quadriga2.log
jdamerow, line 1: Please do not commit this.

quadriga/pom.xml
```diff
@@ -39,11 +39,14 @@
         <mongo.database.authdb>admin</mongo.database.authdb>
 
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
-        <neo4j.database>quadriga</neo4j.database>
+        <neo4j.database>quadriga2</neo4j.database>
```
jdamerow: please don't change this (database url above is fine as it would just set a sensible default, but this should not be changed).

quadriga/pom.xml
```diff
@@ -39,11 +39,14 @@
...
         
-        <citesphere.base.url></citesphere.base.url>
-        <citesphere.client.id></citesphere.client.id>
-        <citesphere.client.secret></citesphere.client.secret>
+        <citesphere.base.url>https://diging-dev.asu.edu/citesphere-review/</citesphere.base.url>
+        <citesphere.client.id>OAUTHCLIENT40</citesphere.client.id>
+        <citesphere.client.secret>1fef2d29-7d30-435e-9210-c56eb18131cf</citesphere.client.secret>
```
jdamerow: these are sensitive information and never be pushed to a public repo!

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
```diff
@@ -132,6 +140,25 @@ public boolean matches(Collection arg0) {
...
+    @Test
+    public void test_deleteCollection_archived() throws CollectionNotFoundException, InvalidObjectIdException {
+        String name = "name";
+        String desc = "description";
+        Collection collection = new Collection();
+        ObjectId id = new ObjectId();
+        collection.setId(id);
+        collection.setName(name);
+        collection.setDescription(desc);
+        Collection response = new Collection();
```
jdamerow: if we change the code to do something to the collection after it was save, then this test will fail if you just return an empty collection object. Maybe just return `collection` in line 155?

jdamerow (changes requested):

quadriga/logs/quadriga2.log
jdamerow, line 1: should not be committed/pushed

quadriga/pom.xml
jdamerow: ?

quadriga/pom.xml
```diff
@@ -44,6 +44,9 @@
         <citesphere.base.url></citesphere.base.url>
         <citesphere.client.id></citesphere.client.id>
         <citesphere.client.secret></citesphere.client.secret>
+        <!--The base.url can have trailing '/' -->
+        <conceptpower.base.url>https://chps.asu.edu/conceptpower/rest/</conceptpower.base.url>
+        <conceptpower.id.endpoint>https://chps.asu.edu/conceptpower/</conceptpower.id.endpoint>
```
jdamerow: this does not belong to this story, does it?

quadriga/src/test/java/edu/asu/diging/quadriga/core/service/impl/CollectionManagerImplTest.java
jdamerow: is this still used?

quadriga/src/main/java/edu/asu/diging/quadriga/web/ListArchivedCollectionController.java
```diff
@@ -0,0 +1,48 @@
...
+
+import edu.asu.diging.quadriga.core.data.CollectionRepository;
+
+@Controller
+public class ListArchivedCollectionController {
+
+    private Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Autowired
+    private CollectionRepository collectionRepo;
```
jdamerow: controllers should not directly talk to the repositories; there should be a service layer in between.

jdamerow (changes requested):

jdamerow (changes requested):
log file is still pushed.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
The collections do not show up (neither on the collections page, nor on the archived one). I ran the mongodb command to update the archived property.

jdamerow:
This is the current data:
```
{ "_id" : ObjectId("61f44cdd3e937c739b167ee9"), "name" : "First collection", "description" : "collection", "apps" : [ "xxx", "x", "xx" ], "_class" : "edu.asu.diging.quadriga.core.model.Collection", "isArchived" : false }
{ "_id" : ObjectId("62d80e7eb68e63665204a8b9"), "name" : "to be archived", "description" : "test", "creationTime" : "2022-07-20T14:17:34.034522Z", "isArchived" : false, "apps" : [ "xx" ], "_class" : "edu.asu.diging.quadriga.core.model.Collection" }
```

ShivaniYerram:
It's working for me, the variable name has changed to "archived" from "isArchived", your data shows "isArchived" maybe that is the issue.

jdamerow:
Make it so, Jenkins.

PR #33, opened 2023-05-25
Q20-15 - story/Q20-15

ShivaniYerram:

The endpoint needs to accept a list of mappings and patterns (pattern1 -> mapping1, pattern2 -> mapping2, ...). Quadriga would then create a new "Collection Mapping" that transforms all the graphs that are matched by one of the patterns with the corresponding pattern. Basically, new groups of triples are being created (in addition to the default mapping one).
  
... describe here ...


jdamerow:
Codefactor failed.
Resolve conflicts please

jdamerow:
Resolve conflicts please

quadriga/pom.xml
```diff
@@ -41,10 +42,19 @@
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
         <neo4j.database>quadriga</neo4j.database>
         
-        <citesphere.base.url></citesphere.base.url>
+        <quadriga.base.url></quadriga.base.url>
+        <quadriga.jobstatus.api></quadriga.jobstatus.api>
+        <quadriga.collectionpage.url></quadriga.collectionpage.url>
+        
+        <citesphere.base.url>https://diging-dev.asu.edu/citesphere-review/</citesphere.base.url>
```
jdamerow: should not be changed

quadriga/pom.xml
```diff
@@ -41,10 +42,19 @@
...
+        <citesphere.client.id></citesphere.client.id>
+        <citesphere.client.secret></citesphere.client.secret>
```
jdamerow, line 50: duplicate

quadriga/pom.xml
```diff
@@ -41,10 +42,19 @@
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
         <neo4j.database>quadriga</neo4j.database>
         
-        <citesphere.base.url></citesphere.base.url>
+        <quadriga.base.url></quadriga.base.url>
+        <quadriga.jobstatus.api></quadriga.jobstatus.api>
+        <quadriga.collectionpage.url></quadriga.collectionpage.url>
```
jdamerow: I'm confused about these two (line 46 and 47). These are path from Quadriga itself, right? Just create a constant in the controller with that path and use the constant when you need it.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/JobStatusController.java
```diff
@@ -0,0 +1,32 @@
...
+    
+    @Autowired
+    private JobManager jobManager;
+
+    
+    @GetMapping(value = "/api/v1/job/{jobId}/status")
+    public ResponseEntity<Job> getJobStatus(@PathVariable String jobId) {
+        Job job = jobManager.get(jobId);
+        if(job == null)
+        {
```
jdamerow: please follow our code style

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/JobStatusController.java
```diff
@@ -0,0 +1,32 @@
...
+            return new ResponseEntity<>(null,HttpStatus.BAD_REQUEST);
```
jdamerow: should return a not found

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,100 @@
...
+            @RequestBody PatternMappingList patternMappingList) {
+
+        List<EventGraph> eventGraphs = eventGraphService.getEventGraphs(new ObjectId(collectionId));
+        if (eventGraphs == null) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+
+        List<JobPatternInfo> jobInfos = new ArrayList<>();
+        for (PatternMapping pattern : patternMappingList.getPatternMappings()) {
+            String jobId = jobManager.createJob(collectionId, pattern.getMappedTripleGroupId(), eventGraphs.size());
```
jdamerow: I think there should be just one job for all transformations. The individual transformations should probalby be horizontally scalable, but there should be just one job to check on.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,100 @@
...
+    @Autowired
+    private AsyncPatternProcessor asyncPatternProcessor;
+
+    @Autowired
+    private EventGraphService eventGraphService;
+
+    @Autowired
+    private JobManager jobManager;
+
+    @PostMapping(value = "/api/v1/collection/{collectionId}/network/map")
```
jdamerow: I think this should just be `/api/v1/collection/{collectionId}/map`

quadriga/src/main/java/edu/asu/diging/quadriga/config/PersistenceConfig.java
```diff
@@ -70,7 +70,7 @@ public PersistenceExceptionTranslationPostProcessor exceptionTranslation() {
     Properties additionalProperties() {
         Properties properties = new Properties();
         properties.setProperty("hibernate.hbm2ddl.auto", "update");
-        properties.setProperty("hibernate.dialect", "org.hibernate.dialect.MySQL5Dialect");
+        properties.setProperty("hibernate.dialect", "org.hibernate.dialect.MySQL8Dialect");
```
jdamerow: should not be changed

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    @Async
+    @Override
+    public void processPattern(String jobId, String collectionId, PatternMapping patternMapping,
+            List<EventGraph> networks) {
+
+        Job job = jobRepository.findById(jobId).orElse(null);
+        PatternCreationEvent patternRoot = patternMapper.mapPattern(patternMapping);
+        
+        if (job == null || patternRoot == null) {
+            job.setStatus(JobStatus.FAILURE);
```
jdamerow: if job is null, this line will throw an exception

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,107 @@
...
+                mappedTripleGroup = mappedTripleGroupService.getByCollectionIdAndId(collectionId,
+                        patternMapping.getMappedTripleGroupId());
+                if (mappedTripleGroup == null) {
+                    job.setStatus(JobStatus.FAILURE);
+                    jobRepository.save(job);
+                    return;
+                }
+            } else {
+                mappedTripleGroup = mappedTripleGroupService.get(collectionId, MappedTripleType.DEFAULT_MAPPING);
+            }
```
jdamerow, line 82: I"m confused what this if/else is for. The would be:
- a list of patterns and mappings is send to Quadriga
- a new mapped triple group is created
- all events that map one of the patterns are transformed according to the submitted mappings
triples should therefore never be added to the default mapping group. and the mapped triple groups that are being created would be of type custom mappings.

jdamerow (changes requested):
There is probably more, but I'll stop here and wait for the next iteration.

ShivaniYerram:
I have raised the PR for further insights from you  regarding the api, mapping type.Kindly let me know.

quadriga/pom.xml
jdamerow: ?

quadriga/pom.xml
jdamerow: this should have a dummy default value

quadriga/pom.xml
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/JobStatusController.java
```diff
@@ -0,0 +1,26 @@
...
+    
+    @Autowired
+    private JobManager jobManager;
+
+    
+    @GetMapping(value = "/api/v1/job/{jobId}/status")
+    public ResponseEntity<Job> getJobStatus(@PathVariable String jobId) {
+        Job job = jobManager.get(jobId);
+        if(job == null){
+            return new ResponseEntity<>(null,HttpStatus.NOT_FOUND);
```
jdamerow: I'm fairly certain there is a constructor that just takes an httpstatus.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+                asyncPatternProcessor.processPattern(jobId, collectionId, pattern, eventGraphs);
+                JobPatternInfo jobInfo = new JobPatternInfo();
+                jobInfo.setJobId(jobId);
+                jobInfo.setTrack(quadrigaBaseUri + quadrigaJobStatusUri + jobId + "/status");
+                jobInfo.setExplore(quadrigaBaseUri + quadrigaCollectionPageUri + collectionId);
+                jobInfos.add(jobInfo);
+            });
+            transformationThreads.add(transformationThread);
+            transformationThread.start();
+        }
```
jdamerow: this above is all business logic and should not be done in the controller

jdamerow: also, you should create threads; use the async functionality of spring instead

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+        for (Thread thread : transformationThreads) {
+            try {
+                thread.join();
+            } catch (InterruptedException e) {
+                System.out.println("Job thread was interrupted due to the exception:"+e);
```
jdamerow: ???

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
```
jdamerow: this is all not necessary when using Spring's async

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+        return new ResponseEntity<>(jobInfos, HttpStatus.OK);
```
jdamerow: There should be just one job for a transformation request. There might be multiple threads started to work on the individual patterns, but there should only be one job.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @GetMapping(value = "/api/v1/job/{jobId}/status")
+    public ResponseEntity<Job> getJobStatus(@PathVariable String jobId) {
```
jdamerow: should be in it's own controller

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+        return new ResponseEntity<>(jobManager.get(jobId), HttpStatus.OK);
+    }
+     
+    class JobPatternInfo {
+        private String jobId;
+        private String track;
```
jdamerow: what is 'track'?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,119 @@
...
+        private String explore;
```
jdamerow: what is 'explore'?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/PatternMapping.java
```diff
@@ -0,0 +1,38 @@
+package edu.asu.diging.quadriga.api.v1.model;
+
+import java.util.List;
+import java.util.Map;
+
+public class PatternMapping {
```
jdamerow: needs javadoc

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/PatternMapping.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private String mappedTripleGroupId;
```
jdamerow, line 16: what does this hold? (when is it filled?)

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/PatternMappingList.java
```diff
@@ -0,0 +1,17 @@
+package edu.asu.diging.quadriga.api.v1.model;
+
+import java.util.List;
+
+public class PatternMappingList {
```
jdamerow: what is this class for?

quadriga/src/main/java/edu/asu/diging/quadriga/config/PersistenceConfig.java
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
```diff
@@ -0,0 +1,33 @@
+package edu.asu.diging.quadriga.core.data;
```
jdamerow: let's put the relational database classes/interfaces into a subpackage `sql` or something like that.

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/events/pattern/PatternAppellationEvent.java
```diff
@@ -0,0 +1,24 @@
+package edu.asu.diging.quadriga.core.model.events.pattern;
```
jdamerow: these are just temporary classes created from the submitted pattern file, right? They should not be in the model package.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
jdamerow: ?

jdamerow (changes requested):
actually I stop here since there seem to be unaddressed comments from before.

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
ShivaniYerram: Made this change in story Q20-16.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
ShivaniYerram: I have created a new constant in MappedTripleType that is CUSTOM_MAPPING.

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
jdamerow: Then this should not be merged like this.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/PatternMappingList.java
jdamerow: Why is this class needed? Can't you just use a list?

quadriga/src/main/java/edu/asu/diging/quadriga/core/exceptions/JobNotFoundException.java
```diff
@@ -0,0 +1,24 @@
+package edu.asu.diging.quadriga.core.exceptions;
+
+public class JobNotFoundException extends Exception {
+    
+
+    private static final long serialVersionUID = -4080713847662867039L;
```
jdamerow: should start at 1

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
jdamerow: no else needed in line 66; if job is null this line will never be reached, so you can just start with an if (not an else if)

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        
+        MappedTripleGroup mappedTripleGroup;
+        try {
+            if (patternMapping.getMappedTripleGroupId() != null && !patternMapping.getMappedTripleGroupId().isEmpty()) {
+                mappedTripleGroup = mappedTripleGroupService.findByCollectionIdAndId(collectionId,
+                        patternMapping.getMappedTripleGroupId());
+                if (mappedTripleGroup == null) {
+                    mappedTripleGroup = new MappedTripleGroup();
+                    mappedTripleGroup.set_id(new ObjectId(patternMapping.getMappedTripleGroupId()));
+                    job.setStatus(JobStatus.PROCESSING);
```
jdamerow: this has already been set in line 71, hasn't it?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,114 @@
...
+                    jobRepository.save(job);
+                }
+            } else {
+                mappedTripleGroup = mappedTripleGroupService.get(collectionId, MappedTripleType.CUSTOM_MAPPING);
```
jdamerow: There can be more than one custom mapping. Get does not create a new mapping, does it?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        } catch (InvalidObjectIdException | CollectionNotFoundException e) {
+            logger.error("Invalid triple group id {} or collection id {} for job {}",
+                    patternMapping.getMappedTripleGroupId(), collectionId, job.getId(), e);
+            System.out.println("Failed 4");
```
jdamerow: remove

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            jobRepository.save(job);
+            return;
+        }
+
+        for (EventGraph network : networks) {
+            List<Graph> extractedGraphs = patternFinder.findGraphsWithPattern(patternMapping.getMetadata(), patternRoot,
+                    network);
+            for (Graph extractedGraph : extractedGraphs) {
+                try {
+                    mappedTripleService.storeMappedGraph(extractedGraph, mappedTripleGroup);
```
jdamerow, line 97: `storeMappedGraph` uses the default mapping to map a graph, which is not what should happen here. The graph should be mapped according to the mapping provided.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/ConceptFinderImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        uri = uri.endsWith("/") ? uri.substring(0, uri.length() - 1) : uri;
+        return uri;
+    }
+
+    private String getNextFormat(String uri) {
+        boolean containsHttps = uri.startsWith("https");
+        boolean endsWithSeparator = uri.endsWith("/");
+
+        if (!endsWithSeparator) {
+            return uri + "/";
```
jdamerow: if it does not end with `/`, don't you still want to fix http with https?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/ConceptFinderImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        uri = uri.replaceFirst("https", "http");
+        uri = uri.endsWith("/") ? uri.substring(0, uri.length() - 1) : uri;
+        return uri;
+    }
+
+    private String getNextFormat(String uri) {
+        boolean containsHttps = uri.startsWith("https");
+        boolean endsWithSeparator = uri.endsWith("/");
+
+        if (!endsWithSeparator) {
```
jdamerow: since you never use `endsWithSeparator` again, just do the `endWith` call right here.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/ConceptFinderImpl.java
```diff
@@ -0,0 +1,67 @@
...
+            return uri + "/";
+        } else if (!containsHttps) {
```
jdamerow: same, just do the `uri.startsWith("https")` call here

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    
+    private final String quadrigaBaseUri = "http://localhost:8081/quadriga/" ;
+    private final String quadrigaJobStatusUri = "api/v1/job/";
+    private final String quadrigaCollectionPageUri = "auth/collections/";
+    
+    @Autowired
+    private AsyncPatternProcessor asyncPatternProcessor;
+
+    @Override
+    public List<JobPatternInfo> getJobPatternInfo(String collectionId, String jobId,List<EventGraph> eventGraphs, PatternMappingList patternMappingList) {
```
jdamerow: this method starts the whole mapping process, doesn't it? It should not be called "get...". That implies it just gets some info, not that it starts a whole process.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/PatternFinderImpl.java
```diff
@@ -0,0 +1,176 @@
...
+import edu.asu.diging.quadriga.core.model.events.CreationEvent;
+import edu.asu.diging.quadriga.core.model.events.RelationEvent;
+import edu.asu.diging.quadriga.core.pattern.PatternAppellationEvent;
+import edu.asu.diging.quadriga.core.pattern.PatternCreationEvent;
+import edu.asu.diging.quadriga.core.pattern.PatternRelationEvent;
+import edu.asu.diging.quadriga.core.service.ConceptFinder;
+import edu.asu.diging.quadriga.core.service.PatternFinder;
+
+@Service
+public class PatternFinderImpl implements PatternFinder {
```
jdamerow, line 38: this class and its methods needs javadoc

quadriga/src/main/resources/log4j.xml
```diff
@@ -48,7 +48,7 @@
     <root>
         <priority value="warn" />
         <appender-ref ref="console" />
-        <appender-ref ref="FILE" />
+        <!--  <appender-ref ref="FILE" />-->
```
jdamerow: don't change that please

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
ShivaniYerram: It creates a new MappedTripleGroup if we give a new MappedTripleType.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
ShivaniYerram: DefaultMapping is the provided mapping from the api that we want our patterns to be mapped to.
A pattern mapping object has ` private String mappedTripleGroupId;
    private Metadata metadata;
    private Map<String, PatternNodeData> nodes;
    private List<Edge> edges;
    `
where Metadata again has these: `private DefaultMapping defaultMapping;
    private Context context;`

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/ConceptFinderImpl.java
```diff
@@ -0,0 +1,67 @@
...
+            uri = uri.replaceFirst("http", "https");
+            uri = uri.endsWith("/") ? uri.substring(0, uri.length() - 1) : uri;
```
ShivaniYerram: Can check whether we need a https with "/" or without "/", once Q20-16 is merged.

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
ShivaniYerram: Resolved.

quadriga/pom.xml
```diff
@@ -45,11 +45,13 @@
         <neo4j.url>http://neo4j:quadriga@localhost:7474/</neo4j.url>
         <neo4j.database>quadriga</neo4j.database>
         
-        <citesphere.base.url></citesphere.base.url>
+        <citesphere.base.url>/</citesphere.base.url>
```
jdamerow: this should not be here

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,61 @@
...
+    @PostMapping(value = "/api/v1/collection/{collectionId}/map")
+    public ResponseEntity<List<JobPatternInfo>> mapPatternToTriples(@PathVariable String collectionId,
+            @RequestBody List<PatternMapping> patternMappingList) {
+
+        List<EventGraph> eventGraphs = eventGraphService.getEventGraphs(new ObjectId(collectionId));
+        
+        if (eventGraphs == null) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+        System.out.println(patternMappingList);
```
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,61 @@
...
+    private MapGraphToTriple mapGraphToTriple;
+
+    @Autowired
+    private JobManager jobManager;
+
+    @PostMapping(value = "/api/v1/collection/{collectionId}/map")
+    public ResponseEntity<List<JobPatternInfo>> mapPatternToTriples(@PathVariable String collectionId,
+            @RequestBody List<PatternMapping> patternMappingList) {
+
+        List<EventGraph> eventGraphs = eventGraphService.getEventGraphs(new ObjectId(collectionId));
```
jdamerow: these can be a lot of graphs. You don't want to load them all into memory. I don't think you want this done here.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,61 @@
...
+        String jobId = jobManager.createJob(collectionId, mappedTripleGroup.get_id().toString(), eventGraphs.size());
+        
+        List<JobPatternInfo> jobInfos = mapGraphToTriple.mapPatterns(collectionId, jobId, eventGraphs, patternMappingList);
+        
+        if(!jobInfos.isEmpty()) {
+            return new ResponseEntity<>(jobInfos, HttpStatus.OK);
+            
+        }
+        
+        return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
jdamerow: when would that be the case? And why does it return "not found"? what is not found?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/JobPatternInfo.java
```diff
@@ -0,0 +1,38 @@
+package edu.asu.diging.quadriga.api.v1.model;
+
+/*
+ * @JobPatternInfo The object of this class is returned when @MapGraphToTripleController is called to map patterns in a collection.
+ * @param jobId is the jobId of the job created for processing the pattern.
+ * @param track is the URI of the status of the job.
+ * @param explore is the URI of the collection whose graphs are being mapped with the input patterns.
+ */
+
```
jdamerow: remove

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/JobPatternInfo.java
```diff
@@ -0,0 +1,38 @@
+package edu.asu.diging.quadriga.api.v1.model;
+
+/*
+ * @JobPatternInfo The object of this class is returned when @MapGraphToTripleController is called to map patterns in a collection.
```
jdamerow: This is the JobPatternInfo class, isn't it? So no need to link to it.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/JobPatternInfo.java
```diff
@@ -0,0 +1,38 @@
...
+ * @param jobId is the jobId of the job created for processing the pattern.
+ * @param track is the URI of the status of the job.
+ * @param explore is the URI of the collection whose graphs are being mapped with the input patterns.
```
jdamerow: I don't think class javadoc as params. Variable explanations should go with the variables or their getters/setters.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/JobPatternInfo.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+    public String getExplore() {
+        return explore;
+    }
+    public void setExplore(String explore) {
+        this.explore = explore;
+    }
+    
+
+
```
jdamerow: why are there so many empty lines?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import edu.asu.diging.quadriga.core.exceptions.JobNotFoundException;
+import edu.asu.diging.quadriga.core.model.EventGraph;
+import edu.asu.diging.quadriga.core.service.AsyncPatternProcessor;
+
+@Service
+public class MapGraphToTripleImpl implements MapGraphToTriple{
+    
+    private final String quadrigaBaseUri = "http://localhost:8081/quadriga/" ;
+    private final String quadrigaJobStatusUri = "api/v1/job/";
+    private final String quadrigaCollectionPageUri = "auth/collections/";
```
jdamerow: What are these doing here? Those should not be hard-coded!

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,49 @@
...
+                return null;
+            }
+            JobPatternInfo jobInfo = new JobPatternInfo();
+            jobInfo.setJobId(jobId);
+            jobInfo.setTrack(quadrigaBaseUri + quadrigaJobStatusUri + jobId + "/status");
+            jobInfo.setExplore(quadrigaBaseUri + quadrigaCollectionPageUri + collectionId);
+            jobInfos.add(jobInfo);
+        }
+        
+        return jobInfos;
```
jdamerow: there should be one job per request, even if there are multiple threads doing the processing of patterns, but the user only cares about all of it being done.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    public List<JobPatternInfo> mapPatterns(String collectionId, String jobId,List<EventGraph> eventGraphs, List<PatternMapping> patternMappingList) {
+        
+        List<JobPatternInfo> jobInfos = new ArrayList<>();
+        
+        for (PatternMapping pattern : patternMappingList) {
+            
+            try {
+                asyncPatternProcessor.processPattern(jobId, collectionId, pattern, eventGraphs);
+            } catch (JobNotFoundException e) {
+                return null;
```
jdamerow: why would that return null? What's the use case you're trying to handle here? Also, if an exception is handled it needs to be logged!

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
jdamerow: where?

quadriga/src/main/java/edu/asu/diging/quadriga/core/model/jobs/JobStatus.java
```diff
@@ -0,0 +1,9 @@
+package edu.asu.diging.quadriga.core.model.jobs;
+
+public enum JobStatus {
```
jdamerow, line 6: javadoc

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
jdamerow: That does not make sense. There can be multiple custom mappings for a collection, each in their own custom mapping triple group. Also, a method called `get` should get an element, not create it (I missed that before).

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+            logger.error("Invalid triple group id {} or collection id {} for job {}",
+                    patternMapping.getMappedTripleGroupId(), collectionId, job.getId(), e);
+
+            job.setStatus(JobStatus.FAILURE);
+            jobRepository.save(job);
+            return;
+        }
+
+        for (EventGraph network : networks) {
+            List<Graph> extractedGraphs = patternFinder.findGraphsWithPattern(patternMapping.getMetadata(), patternRoot,
```
jdamerow: is that method only finding graphs or also mapping them? It seems like it's also mapping them, in which case the method name is incorrect.

quadriga/src/test/java/edu/asu/diging/quadriga/api/v1/AddNetworkApiControllerTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @InjectMocks
+    private AddNetworkApiController addNetworkApiController;
+
+    @BeforeEach
+    public void setUp() {
+        MockitoAnnotations.openMocks(this);        
+    }
+
+    @Test
+    void testProcessJson_Success() throws NodeNotFoundException, InvalidObjectIdException, CollectionNotFoundException {
```
jdamerow: please adhere to our test naming guidelines

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
ShivaniYerram: There is just a single job doing this as there is only 1 jobId for all the jobs. JobPatternInfo gives us the number of patterns processed and their information for tracking and exploring with the help of jobId.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
ShivaniYerram: Sorry, there was a mistake,  the function 'get' creates a new mappedTripleGroup if there isn't one for the given collection ID and MappedTripleType.

quadriga/pom.xml
jdamerow: unaddressed

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,66 @@
...
+            @RequestBody List<PatternMapping> patternMappingList) {
+        
+        
+        int pageNumber = 0; // Set initial page number
+        int pageSize = 10;
+
+        List<EventGraph> eventGraphs = eventGraphService.getEventGraphsByCollectionId(new ObjectId(collectionId), PageRequest.of(pageNumber, pageSize));
+        
+        if (eventGraphs == null) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
jdamerow: I think the logic should rather be to check if the collection exists, not the if event graphs exists. it's fine (even if it doesn't make much sense) to try to run mappings on 0 event graphs. An invalid collection id, however, should return a not found.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+               
+        MappedTripleGroup mappedTripleGroup = new MappedTripleGroup();
+        mappedTripleGroup.set_id(new ObjectId());
+        mappedTripleGroup.setCollectionId(new ObjectId(collectionId));
```
jdamerow: this should not be part of the controller, that's business logic.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/MapGraphToTripleController.java
```diff
@@ -0,0 +1,66 @@
...
+        
+        String jobId = jobManager.createJob(collectionId, mappedTripleGroup.get_id().toString(), eventGraphs.size());
+        
+        List<JobPatternInfo> jobInfos = mapGraphToTriple.mapPatterns(collectionId, jobId, eventGraphs, patternMappingList);
```
jdamerow: same with these two lines. The controller should just call one method that takes care of the rest.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/model/JobPatternInfo.java
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import edu.asu.diging.quadriga.core.exceptions.JobNotFoundException;
+import edu.asu.diging.quadriga.core.model.EventGraph;
+import edu.asu.diging.quadriga.core.service.AsyncPatternProcessor;
+
+@Service
+public class MapGraphToTripleImpl implements MapGraphToTriple{
+    
+//    private final String quadrigaBaseUri = "http://localhost:8081/quadriga/" ;
+//    private final String quadrigaJobStatusUri = "api/v1/job/";
+//    private final String quadrigaCollectionPageUri = "auth/collections/";
```
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+import edu.asu.diging.quadriga.api.v1.model.JobPatternInfo;
+import edu.asu.diging.quadriga.api.v1.model.PatternMapping;
+import edu.asu.diging.quadriga.api.v1.service.MapGraphToTriple;
+import edu.asu.diging.quadriga.core.exceptions.JobNotFoundException;
+import edu.asu.diging.quadriga.core.model.EventGraph;
+import edu.asu.diging.quadriga.core.service.AsyncPatternProcessor;
+
+@Service
+public class MapGraphToTripleImpl implements MapGraphToTriple{
```
jdamerow: bad class name. class names should be nouns.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/MapGraphToTripleImpl.java
jdamerow: ah, right. I think I would structure the response differently though. multiple job infos with the same job id indicates the ids could be different. Instead, i would have one job object with an id that then has a list of mapping infos.

quadriga/src/main/java/edu/asu/diging/quadriga/core/data/IdGenerator.java
jdamerow: ?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+            if (patternMapping.getMappedTripleGroupId() != null && !patternMapping.getMappedTripleGroupId().isEmpty()) {
+                mappedTripleGroup = mappedTripleGroupService.findByCollectionIdAndId(collectionId,
+                        patternMapping.getMappedTripleGroupId());
+                if (mappedTripleGroup == null) {
+                    mappedTripleGroup = new MappedTripleGroup();
+                    mappedTripleGroup.set_id(new ObjectId(patternMapping.getMappedTripleGroupId()));
+                    jobRepository.save(job);
+                }
+            } else {
+                mappedTripleGroup = mappedTripleGroupService.getMappedTripleGroupIfExistsOrAdd(collectionId, MappedTripleType.CUSTOM_MAPPING);
```
jdamerow, line 81: this does not seem right, wouldn't that always return the first custom mapping it finds?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
jdamerow: there needs to be some documentation about this somewhere, or even better, default mapping should not be called default mapping.

jdamerow (changes requested):

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/GraphTripleMapperController.java
```diff
@@ -0,0 +1,56 @@
...
+    public ResponseEntity<List<JobPatternInfo>> mapPatternToTriples(@PathVariable String collectionId,
+            @RequestBody List<PatternMapping> patternMappingList) throws InvalidObjectIdException, CollectionNotFoundException {
+               
+        int pageNumber = 0; // Set initial page number
+        int pageSize = 10;
+        if(collectionManager.findCollection(collectionId) == null) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+
+        List<EventGraph> eventGraphs = eventGraphService.getEventGraphsByCollectionId(new ObjectId(collectionId), PageRequest.of(pageNumber, pageSize));
```
jdamerow, line 46: does this mean only the first 10 graphs are being mapped? that doesn't make sense.

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/GraphTripleMapperImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import edu.asu.diging.quadriga.core.exceptions.JobNotFoundException;
+import edu.asu.diging.quadriga.core.model.EventGraph;
+import edu.asu.diging.quadriga.core.model.MappedTripleGroup;
+import edu.asu.diging.quadriga.core.service.AsyncPatternProcessor;
+import edu.asu.diging.quadriga.core.service.JobManager;
+import edu.asu.diging.quadriga.core.service.MappedTripleGroupService;
+
+@Service
+@PropertySource("classpath:config.properties")
+public class GraphTripleMapperImpl implements GraphTripleMapper{
```
jdamerow, line 28: this is business logic and should not be in the api package.

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
jdamerow: this still

quadriga/src/main/java/edu/asu/diging/quadriga/api/v1/service/impl/GraphTripleMapperImpl.java
```diff
@@ -0,0 +1,74 @@
...
+                asyncPatternProcessor.processPattern(jobId, collectionId, pattern, eventGraphs);
+            } catch (JobNotFoundException e) {
+                logger.error("Job not found for Job id "+jobId+" and collection Id "+collectionId,e);
+                return jobInfos;
+            }
+            JobPatternInfo jobInfo = new JobPatternInfo();
+            jobInfo.setJobId(jobId);
+            jobInfo.setTrack(quadrigaBaseUri + quadrigaJobStatusUri + "/" + jobId + "/status");
+            jobInfo.setExplore(quadrigaBaseUri + quadrigaCollectionPageUri + collectionId);
+            jobInfos.add(jobInfo);
```
jdamerow, line 69: I'm confused, don't all job infos look the same? there is just one job id, isn't there? and all the other variables are also the same? why does this need to be a list?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/EventGraphService.java
```diff
@@ -26,6 +28,13 @@ public interface EventGraphService {
      */
     public EventGraph findLatestEventGraphByCollectionId(ObjectId collectionId);
     
+    /**
+     * Retrieves all eventGraphs mapped to a collection
```
jdamerow, line 32: does this mean "belong to a collection"?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/EventGraphService.java
```diff
@@ -39,4 +48,14 @@ public interface EventGraphService {
      * @param collectionId
      */
     public void mapNetworkAndSave(Graph graph, String collectionId);
+    
+    /**
+     * Finds all eventGraphs mapped to a collection in the descending order of creation time
```
jdamerow, line 53: same here, "belonging"?

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/AsyncPatternProcessorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    @Async
+    public void processPattern(String jobId, String collectionId, PatternMapping patternMapping,
+            List<EventGraph> networks) throws JobNotFoundException {
+
+        Job job = jobRepository.findById(jobId).orElse(null);
+        PatternCreationEvent patternRoot = patternMapper.mapPattern(patternMapping);
+        
+        if (job == null) {
+            throw new JobNotFoundException();  
+        }
```
jdamerow, line 65: this should be done before line 61

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/PatternMapperImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    /* (non-Javadoc)
+     * @see edu.asu.diging.quadriga.core.service.PatternMapper#mapPattern(edu.asu.diging.quadriga.api.v1.model.PatternMapping)
+     */
+    @Override
+    public PatternCreationEvent mapPattern(PatternMapping patternMapping) {
+        List<Entry<String, PatternNodeData>> startNodes = patternMapping.getNodes().entrySet().stream().filter(
+                node -> patternMapping.getEdges().stream().noneMatch(edge -> edge.getTarget().equals(node.getKey())))
+                .collect(Collectors.toList());
+
+        if (startNodes.size() != 1) {
```
jdamerow, line 42: needs a comment why this is

quadriga/src/main/java/edu/asu/diging/quadriga/core/service/impl/PatternMapperImpl.java
```diff
@@ -0,0 +1,98 @@
...
+            return null;
+        }
+        String key = startNodes.get(0).getKey();
+        PatternNodeData node = startNodes.get(0).getValue();
+        PatternCreationEvent root = createGraph(key, node, patternMapping);
+        return root;
```
jdamerow, line 48: merge with line 47

jdamerow (changes requested):

